PR #2, opened 2021-03-13
AK-14184 Implement data source for policy prefix list

spikewang:
Add data source support for policy prefix list to all user to use already
created policy prefix list.


spikewang:
some changes are due to gofmt... Doc will be re-generated at once when we are doing the next release.

spikewang:
Using it like this:

```
data "alkira_policy_prefix_list" "test1" {
  name = "test1"
}
```

sahilthapar1234 (approved):

PR #3, opened 2021-03-13
AK-14185 Add data resource for group

spikewang:
Add data resource support for group


spikewang:
The documentation will be generated later at once with release. Using it like this:

```
data "alkira_group" "test1" {
  name = "test1"
}
```

sahilthapar1234 (approved):

PR #8, opened 2021-04-21
AK-15200 Add data source support for segment and credential

spikewang:
Add data source support for segment and credential.

+ Update alkira-client-go to pick up new API.
+ Add documentation and examples for data source.
+ Trying out a new release notes doc.


spikewang:
I always wanted to add a release notes... let's try out and see if this works out... sadly, we can't easily to check the generated docs so far (not I know of)... We may explore other options... terraform-plugin-docs seems like stalled for a while now...

sanchit3110 (approved):
Looks good

rvichery (approved):
PR looks good to me!

PR #9, opened 2021-05-03
AK-13448 Add data source for billing tag with examples

spikewang:
Add data source alkira_billing_tag with examples for both resource
and data source. Remove not supported data sources as well to avoid
confusion.


spikewang:
There is new field Description added as well, but it needs update on the alkira-client-go. Will have a separate commit for that.

alkira/resource_alkira_billing_tag.go
```diff
@@ -51,6 +54,17 @@ func resourceBillingTagRead(d *schema.ResourceData, meta interface{}) error {
...
+	log.Printf("[INFO] Billing Tag Updating")
+	err := client.UpdateBillingTag(d.Get("tag_id").(int), d.Get("name").(string))
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+
 	return resourceBillingTagRead(d, meta)
```
rvichery, line 68: This becomes unreachable code. Should not have compiled with this.

alkira/data_source_alkira_billing_tag.go
```diff
@@ -0,0 +1,45 @@
...
+
+		Read: dataSourceAlkiraBillingTagRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "The name of the billing tag.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"billing_tag_id": {
```
rvichery, line 22: Would it be better to use the same attribute name as in `resourceAlkiraBillingTag` with `tag_id` instead of `billing_tag_id`?

rvichery (changes requested):
The change in `resourceBillingTagUpdate` has unreachable code at line 68. Also we should probably use the same attribute for the billing tag id between the resource and datasource.

alkira/data_source_alkira_billing_tag.go
spikewang: we do have other "tag" introduced later on, hence the change was made here... actually, id should be sufficient... `data.alkira_billing_tag.id` is the way we will use this, neither billing_tag_id or `tag_id` is necessary...

alkira/resource_alkira_billing_tag.go
spikewang: compiling works fine. it returns the same actually, but yep, forgot...

rvichery (approved):
Looks good!

PR #10, opened 2021-05-18
AK-13448 Add Description field for Billing Tag

spikewang:
Add support of Description field for Billing Tag.

### DIFF
--- a/alkira/resource_alkira_billing_tag.go
+++ b/alkira/resource_alkira_billing_tag.go
@@ -2,7 +2,6 @@ package alkira
 
 import (
 	"log"
-	"strconv"
 
 	"github.com/alkiranet/alkira-client-go/alkira"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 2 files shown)


spikewang:
Also, switch the ID to from int to string to be consistent with Terraform concept of Id.

alkira/resource_alkira_billing_tag.go
```diff
@@ -11,17 +10,22 @@ import (
...
 				Required:    true,
 			},
+			"description": {
+				Description: "Billing Tag Description.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
 			"tag_id": {
 				Description: "Billing Tag ID.",
 				Type:        schema.TypeInt,
```
rvichery, line 31: Should we keep this one as TypeInt? seems like we are only using it as a string

rvichery (approved):
Looks good to me

alkira/resource_alkira_billing_tag.go
spikewang: I should remove this one now... No need anymore. The default Id should just work..

PR #11, opened 2021-05-18
AK-15710 Convert resource Id from int to string

spikewang:
Convert resources' Id from int to string to consistent with Terraform
convention. This commit includes connector_aws_vpc, group and segment. Also,
gofmt one more time across.


rvichery (approved):
Looks good to me

PR #23, opened 2021-07-22
AK-17112 Fix the disable_advertise_on_prem_routes in connector_ipsc

spikewang:
Fix the disable_advertise_on_prem_routes variable that should be reversed in the
API call.


bchakrav1 (approved):
Looks good

PR #26, opened 2021-08-24
AK-17875 update documentation, rule_id is mandatory for policy rule list

AbhinavkumarL:


docs/resources/policy_rule_list.md
```diff
@@ -35,9 +35,8 @@ description: |-
 Required:
 
 - **priority** (Number)
+- **rule_id** (Number)
```
spikewang, line 38: Please modify the source file and fix the field properly from `required` to `optional` and re-generate the this doc

spikewang (changes requested):

PR #27, opened 2021-08-24
AK-17495 group description is optional

AbhinavkumarL:


docs/resources/group.md
```diff
@@ -20,11 +20,11 @@ Groups can contain one or many connectors across different segments. This groupi
 
 ### Required
 
-- **description** (String) The description of the group.
```
spikewang, line 23: This file is automatically generated... You need to modify the source file and re-generate the doc.

spikewang (changes requested):

PR #32, opened 2021-09-02
AK-17932 Fix service-pan to mark credential_id as required

spikewang:
The credential_id is required var for service-pan.


AbhinavkumarL (approved):

rvichery (approved):
👍

PR #59, opened 2021-12-02
AK-20542 Add resources for lists

spikewang:
Add resources for as-path-list, community-list, extended-community-list and
global-cidr-list.


spikewang:
Tested with create, delete and update.

AbhinavkumarL (approved):

PR #65, opened 2021-12-17
AK-20868 Change subscription_id of credential_azure_vnet as optional

spikewang:
subscription_id of credential_azure_vnet is not required for creating
credential anymore.


rvichery (approved):

wcollins (commented):
Approved

PR #70, opened 2022-02-14
AK-22042 Remove deprecated vars from connector-azure-vnet

spikewang:
Remove vnet_subscription_id and vnet_region from connector-azure-vnet. API
doesn't need them anymore. Also, update alkira-client-go to v0.8.4 for latest
API changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/alkiranet/terraform-provider-alkira
 go 1.16
 
 require (
-	github.com/alkiranet/alkira-client-go v0.8.3
+	github.com/alkiranet/alkira-client-go v0.8.4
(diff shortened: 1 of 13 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/alkiranet/terraform-provider-alkira
 go 1.16
 
 require (
-	github.com/alkiranet/alkira-client-go v0.8.3
+	github.com/alkiranet/alkira-client-go v0.8.4
```
mac-alkira, line 6: Do we have any rules about versioning the alkira client? I just did a minor version when I added a couple of fields. Looks like you did the same.

mac-alkira (approved):

PR #72, opened 2022-02-14
AK-21408 Add enabled field to all connectors

spikewang:
Add new optional "enabled" field to all connnectors. The default value will
be true for all connectors.


spikewang:
I have tested all connectors create/delete with this. The optional field will be true by default.

AbhinavkumarL (approved):

mac-alkira (approved):
Looks good. Did we run an integration test through the terraform test repo for each one of these connectors?

PR #74, opened 2022-02-15
AK-21153 Add optional service_tags to connector_azure_vnet

spikewang:
Add optional new argument service_tags to connector_azure_vnet.

### DIFF
--- a/alkira/resource_alkira_connector_azure_vnet.go
+++ b/alkira/resource_alkira_connector_azure_vnet.go
@@ -74,6 +74,15 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 				Type:        schema.TypeInt,
 				Required:    true,
 			},
+			"service_tags": {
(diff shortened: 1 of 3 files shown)


AbhinavkumarL (approved):

alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -74,6 +74,15 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 				Type:        schema.TypeInt,
 				Required:    true,
 			},
+			"service_tags": {
+				Description: "list of service tags from Azure. Providing a service tag here," +
+					"would result in service tag route configuration on VNET route table, so" +
```
mac-alkira, line 79: You need a space between "so" "that" that'll show up as one word. Yeah it looks like you either need to add a space either at the end of each concatenation or at the beginning of the next line. These words are going to be squished together.

PR #75, opened 2022-02-18
add support for gcp routing options

mac-alkira:


alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -40,6 +40,33 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"prefix_list_ids": {
+							Description: "Ids of prefix lists defined on the network.",
+							Type:        schema.TypeList,
+							Required:    true,
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+						},
+						"route_import_mode": {
+							Description: "The routeImportMode is an instruction which specifies " +
```
spikewang, line 60: Do we need to repeat routeImportMode here? It's kind of redundant and inconsistent with the argument name. How about "This is an..."?

alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -40,6 +40,33 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
...
+								"the source of the routes that need to be imported. Only " +
+								"ADVERTISE_DEFAULT_ROUTE and ADVERTISE_CUSTOM_PREFIX are valid inputs.",
```
spikewang, line 62: we could use markdown to make it clearer. `ADVERTISE_DEFAULT_ROUTE` and `ADVERTISE_CUSTOM_PREFIX`.

spikewang (changes requested):

alkira/resource_alkira_connector_aws_vpc_helper.go
```diff
@@ -0,0 +1,79 @@
+package alkira
```
spikewang, line 1: Thanks man, this is much cleaner!

alkira/resource_alkira_connector_gcp_vpc.go
spikewang: routeImportMode is still here?

alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -40,6 +40,35 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
...
+							Required:    true,
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+						},
+						"custom_prefix": {
+							Description: "The routeImportMode is an instruction which specifies " +
+								"the source of the routes that need to be imported. Only " +
+								"`ADVERTISE_DEFAULT_ROUTE` and `ADVERTISE_CUSTOM_PREFIX` are valid inputs.",
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: validateCustomPrefix,
```
spikewang, line 65: This could be done directly like this:

```
ValidateFunc: validation.StringInSlice([]string{"ADVERTISE_DEFAULT_ROUTE", "ADVERTISE_CUSTOM_PREFIX"}, false),
```

alkira/resource_connector_gcp_vpc_helper.go
```diff
@@ -0,0 +1,64 @@
...
+
+		if v, ok := cfg["custom_prefix"].(string); ok {
+			gcp.ImportOptions.RouteImportMode = v
+		}
+	}
+
+	return gcp
+}
+
+func validateCustomPrefix(v interface{}, p cty.Path) diag.Diagnostics {
```
spikewang, line 44: We don't need this... There is built in validation function

spikewang (changes requested):

spikewang (changes requested):
You will also need to add those two to the `read()`.

PR #76, opened 2022-02-28
AK-22536 Mark asn argument in alkira_segment resource as optional

spikewang:
Mark "asn" argument in alkira_segment as optional with default value of
65514.


AbhinavkumarL (approved):

spikewang:
This won't cause any backward compatibility issues.

PR #79, opened 2022-03-08
AK-22855 Add alkira_group_connector resource and data source

spikewang:
Add new alkira_group_connector resource and data source which is same as
the current alkira_group resource and data source.


alkira/resource_alkira_connector_group.go
```diff
@@ -0,0 +1,91 @@
...
+
+func resourceAlkiraGroupConnector() *schema.Resource {
+	return &schema.Resource{
+		Description: "Manage connector groups\n\n" +
+			"Connector Groups can contain one or many connectors across different segments. " +
+			"This grouping of connectors can be for policy enforcement purposes or " +
+			"for monitoring purposes within the network. It allows for easier policy " +
+			"assignment by assigning policies to the entire group at the same time " +
+			"instead of having to assign them individually.",
+		Create: resourceGroupConnector,
```
mac-alkira, line 18: Nit: our current convention is to name this as `resource<name>Create`

mac-alkira (approved):

alkira/resource_alkira_connector_group.go
mac-alkira: I checked a few other places in our repo. I'm the one who isn't keeping with convention by adding Create at the end of the function name.

PR #80, opened 2022-03-08
AK-22856 Add alkira_group_user resource and data source

spikewang:
Add alkira_group_user resource and data source to support Alkira User Group.


spikewang:
Tests were added to the test repo.

mac-alkira (approved):

PR #82, opened 2022-03-09
AK-22776 Add new resource alkira_segment_resource

spikewang:
Add new resource alkira_segment_resource.


spikewang:
will update the example when confirming one detail

mac-alkira (approved):

PR #83, opened 2022-03-09
Bug/AK-22937 : Fix for NAT policy rules inputs being read incorrectly from terraform config file

mac-alkira:


spikewang (approved):
Looks good! Good catch

PR #92, opened 2022-03-14
AK-22969 Add resource alkira_connector_akamai_prolexic

spikewang:
Add the initial support for alkira_connector_akamai_prolexic. This feature is still under active development and this is only for BETA testing so far.


alkira/resource_alkira_connector_akamai_prolexic.go
```diff
@@ -0,0 +1,256 @@
...
+							Required:    true,
+						},
+						"enable_route_advertisement": {
+							Description: "Whether enabling route advertisement.",
+							Type:        schema.TypeBool,
+							Required:    true,
+						},
+					},
+				},
+				Required: true,
```
mac-alkira, line 57: Stylistically we should put the Required/Optional field above Elem. To be clearer, in the case where Elem is a type or set we should put the required field above Elem. Simply for the sake of readability. Not a deal breaker. Just a thought.

alkira/resource_alkira_connector_akamai_prolexic.go
```diff
@@ -0,0 +1,256 @@
...
+										Required:    true,
+									},
+									"akamai_overlay_tunnel_ip": {
+										Description: "The overlay IP of the GRE tunnel on the Alkira side.",
+										Type:        schema.TypeString,
+										Required:    true,
+									},
+								},
+							},
+							Required: true,
```
mac-alkira, line 128: See above.

alkira/resource_alkira_connector_akamai_prolexic.go
```diff
@@ -0,0 +1,256 @@
...
+						},
+					},
+				},
+				Required: true,
```
mac-alkira, line 132: See above.

alkira/resource_alkira_connector_akamai_prolexic.go
```diff
@@ -0,0 +1,256 @@
...
+
+	d.Set("akamai_bgp_asn", connector.AkamaiBgpAsn)
+	d.Set("billing_tag_ids", connector.BillingTags)
+	d.Set("cxp", connector.CXP)
+	d.Set("group", connector.Group)
+	d.Set("enabled", connector.Enabled)
+	d.Set("name", connector.Name)
+	d.Set("size", connector.Size)
+
+	if len(connector.Segments) > 0 {
```
mac-alkira, line 175: This check should be placed nearer the top of the function. No need to set all those other fields if the check for connector segment fails. Not a big deal here but a solid practice nonetheless.

alkira/resource_alkira_connector_akamai_prolexic_helper.go
```diff
@@ -0,0 +1,80 @@
+package alkira
+
+import (
+	"log"
+
+	"github.com/alkiranet/alkira-client-go/alkira"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func expandConnectorAkamaiTunnelIps(in *schema.Set) []alkira.ConnectorAkamaiProlexicTunnelIp {
```
mac-alkira, line 10: We really should have tests for these "expand" type helper functions. The NAT policy rule bug we ran into could've been avoided by just checking that fields were being read in correctly.

mac-alkira (approved):

alkira/resource_alkira_connector_akamai_prolexic_helper.go
spikewang: I did. Every single field was tested through TF configs, NAT policy simply is simply not tested enough.

alkira/resource_alkira_connector_akamai_prolexic.go
spikewang: No, even segment is not valid. We still want to set the fields as much as possible, eventually, it will produce a diff.. The whole point of read is to generate a diff as much as we could.

PR #98, opened 2022-04-06
AK-23526 Change panorama_ip_address to panorama_ip_addresses in servi…

spikewang:
…ce_pan

Change panorama_ip_address to panorama_ip_addresses to support multiple IP
addresses and change global_protect to optional with default to false to
maintain backward compatiblity with old configs.


spikewang:
Just a preview and needs to wait for the new version of alkira-client-go to merge. Tested with configs (create/update/etc).

mac-alkira (approved):

kenkkim (approved):

PR #99, opened 2022-04-06
AK-23527 Add customer_asn field to connector_cisco_sdwan

spikewang:
Add customer_asn field to connector_cisco_sdwan and move alkira-client-go to
v0.9.6 to pick up the changes.


spikewang:
This is optional new field and I could verify that the old config will still work without specifying it.

mac-alkira (approved):

PR #100, opened 2022-04-07
AK-23528 Add new tgw_attachment block to connector_aws_vpc

spikewang:
Add new optional tgw_attachment block to connector_aws_vpc.


spikewang:
A simple block will be like this:

```
tgw_attachment {
    subnet_id = aws_subnet.tftest4_subnet2.id
    az        = "us-west-2a"
  }
```

Tested on staging cluster with old config, update, and new config.

alkira/resource_alkira_connector_aws_vpc_helper.go
```diff
@@ -77,3 +77,27 @@ func generateUserInputPrefixes(cidr []interface{}, subnets *schema.Set) ([]alkir
 
 	return prefixes, nil
 }
+
+// expandAwsVpcTgwAttachments expand tgw_attachment of connector_aws_vpc
+func expandAwsVpcTgwAttachments(in *schema.Set) []alkira.TgwAttachment {
```
mac-alkira, line 82: I think these expand helper functions merit unit tests. It's technically a simple conversion, but mistakes happen and the test is easy to write.

mac-alkira (approved):

PR #106, opened 2022-05-02
AK-23532 Add new routing options support to connector_azure_vnet

spikewang:
Refactoring the current connector_azure_vnet to support CIDR and subnet level routing options. By using `vnet_cidr` and `vnet_subnet`, user could specify `service_tags`, `prefix_list` with routing mode.

+ Add new block `vnet_cidr`.
+ Add new block `vnet_subnet`.
+ Update `size` for the connector.


spikewang:
For examples. please refer to use internal test repo. I have updated configurations with new tests.

alkira/resource_alkira_connector_azure_vnet_helper.go
```diff
@@ -0,0 +1,97 @@
...
+	serviceRoutes := alkira.ConnectorVnetServiceRoutes{}
+
+	// Processing vnet_subnet blocks
+	for _, block := range d.Get("vnet_subnet").(*schema.Set).List() {
+		content := block.(map[string]interface{})
+
+		if _, ok := content["routing_options"].(string); ok {
+			subnetImportOption := alkira.ConnectorVnetImportOptionsSubnet{}
+
+			if v, ok := content["subnet_id"].(string); ok {
```
mac-alkira, line 24: Just thinking out loud here. I found out that if `content["routing_options"]` is not a string type ok evaluates to false. We are not notified that there was a failure in our type assertion. The key could exist and its associated value, but the program continues to run as if it did not. I've been thinking about this and researching and I haven't found a better solution. Something to think about for us. This behavior caused a bug for me that testing revealed.

mac-alkira (approved):

alkira/resource_alkira_connector_azure_vnet_helper.go
spikewang: Yep, so for now, testing is the best we could do to catch this kinds of case… so I made bunch of new configs.

Even it failed, API should have the proper response no matter what….

PR #109, opened 2022-05-03
AK-24316 Allow ENV variable to set alkira_credential_azure_vnet

spikewang:
Add new variables to allow add credentials from ENV variables.

### DIFF
--- a/alkira/resource_alkira_credential_azure_vnet.go
+++ b/alkira/resource_alkira_credential_azure_vnet.go
@@ -9,6 +9,10 @@ import (
 
 func resourceAlkiraCredentialAzureVnet() *schema.Resource {
 	return &schema.Resource{
+		Description: "Manage AWS credential for authentication.\n\n" +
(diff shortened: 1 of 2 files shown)


alkira/resource_alkira_credential_azure_vnet.go
```diff
@@ -24,21 +28,33 @@ func resourceAlkiraCredentialAzureVnet() *schema.Resource {
 				Description: "The Application ID",
 				Type:        schema.TypeString,
 				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc(
```
mac-alkira, line 31: Why are these parameters on different lines? It shouldn't be too long of a function on the same line.

mac-alkira (approved):

alkira/resource_alkira_credential_azure_vnet.go
spikewang: follow the same convention as other providers... gofmt doesn't pick it and then I am okey with it.

PR #110, opened 2022-05-03
AK-24343 Add ENV variable support for specifying credential_oci_vcn

spikewang:
Add ENV variable support for specifying credential_oci_vcn and update
aws_vpc and azure_vnet documentation.

### DIFF
--- a/alkira/resource_alkira_credential_azure_vnet.go
+++ b/alkira/resource_alkira_credential_azure_vnet.go
@@ -9,10 +9,11 @@ import (
 
 func resourceAlkiraCredentialAzureVnet() *schema.Resource {
 	return &schema.Resource{
-		Description: "Manage AWS credential for authentication.\n\n" +
-			"You can provide your credentials via the `AK_AZURE_APPLICATION_ID` " +
-			"`AK_AZURE_SUBSCRIPTION_ID`, `AK_AZURE_SECRET_KEY` and " +
-			"`AK_AZURE_TENANT_ID` environment variables.",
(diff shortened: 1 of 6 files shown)


alkira/resource_alkira_credential_azure_vnet.go
```diff
@@ -9,47 +9,48 @@ import (
...
+			"* AK_AZURE_TENANT_ID\n",
 		Create: resourceCredentialAzureVnet,
 		Read:   resourceCredentialAzureVnetRead,
 		Update: resourceCredentialAzureVnetUpdate,
 		Delete: resourceCredentialAzureVnetDelete,
 
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
-				Description: "The name of the credential",
+				Description: "The name of the credential.",
```
mac-alkira, line 24: This is a little thing but there is inconsistency here. Some of the fields include 'The' as the first word in the descriptor. We either need it to be 'The' for all of them or in this case 'Name of the credential.' I don't particularly care which though I think removing 'The' is more succinct and less repetitive. So yes, lets remove 'The' as a general rule.

mac-alkira: Oh hahaha I just saw you did that below. Sorry. You just missed this 'The'.

mac-alkira (approved):

alkira/resource_alkira_credential_azure_vnet.go
spikewang: Good point! Hehe, I was puzzled too why `The`.

PR #113, opened 2022-05-11
AK-24457 Set export routing field in connector_azure_vnet

spikewang:
Set the export routing options in connector_azure_vnet when vnet_cidr
or vnet_subnet block was used.


spikewang:
Notice that, currently, to use this, you will need the optional field `subscription_id` in the `credential_azure_vnet`. It's being worked on and should be fixed in the next release.

mac-alkira (approved):

PR #123, opened 2022-06-09
AK-25315: update ServicePanInstance to include id when updating pan

kenkkim:
https://alkiranet.atlassian.net/browse/AK-25317
PAN instance was getting destroyed and recreated when id was omitted.


alkira/resource_alkira_service_pan.go
```diff
@@ -97,6 +97,11 @@ func resourceAlkiraServicePan() *schema.Resource {
 							Type:        schema.TypeString,
 							Required:    true,
 						},
+						"id": {
+							Description: "The id of the PAN instance.",
```
spikewang, line 101: For description, we always do ID, all uppercase

spikewang:
Don't forget to generate the doc for service-pan

spikewang (approved):
Thanks!

PR #134, opened 2022-08-01
AK-26737 - update cidr field to cidrs list in segment resource

mac-alkira:


alkira/resource_alkira_segment.go
```diff
@@ -116,12 +116,34 @@ func resourceSegmentDelete(d *schema.ResourceData, m interface{}) error {
...
-		Name:    d.Get("name").(string),
+		Asn:  d.Get("asn").(int),
+		Name: d.Get("name").(string),
 		ReservePublicIPsForUserAndSiteConnectivity: d.Get("reserve_public_ips").(bool),
 	}
 
+	// User may include 1 or more cidrs. Set them accordingly
+	cidrs := convertTypeListToStringList(d.Get("cidrs").([]interface{}))
+
+	if len(cidrs) == 1 {
```
spikewang, line 129: Why do we need this? I thought `IpBlock` was deprecated... We only need to care `IpBlocks`... nope?

spikewang:
@mac-alkira Could we also update the example for segment?

spikewang (approved):
Looks good!

PR #142, opened 2022-08-05
AK-24147 Add resource for routing policy

spikewang:
Add the Initial support for routing policy.


spikewang:
This is only the initial support and it may still have quirks here and there. I think it's good to call it a BETA feature and we could start from there.

spikewang:
Hold on this for now. I did several more experiments. It's probably good idea to flatten out the `rule` block. Those nested blocks are pointless.

spikewang:
Done, flattened out. The new config looks much cleaner now.

wcollins (approved):

PR #143, opened 2022-08-10
AK-27041 Switch endpoint of connector_ipsec to TypeList

spikewang:
Switch endpoint block of connector_ipsec to TypeList to avoid generating uncessary
diff with TypeSet.


spikewang:
The similar issue was noticed with resources too. It's good idea to simply switch to use TypeList instead of TypeSet... I have tested various update case to verify the fix.

spikewang:
The issue was described and discussed detailedly in this thread: https://discuss.hashicorp.com/t/typeset-is-not-picking-up-correct-changes/23735/2.

kenkkim (approved):

PR #145, opened 2022-08-10
AK-26615 Add IPv6 support to alkira_internet_application

spikewang:
+ Add new `internet_protocol` to alkira_internet_application to select used protocol.
+ Add new arguments to segment to allow translation of IPv6 to IPv4 and specify source IPv4 pool.


spikewang:
This whole thing is behind the feature flag for now. I validate the change won't break any current functionalities as well.

mac-alkira (approved):

PR #148, opened 2022-08-12
AK-27133 Update segment resource to emit entire pool list section if …

spikewang:
…not used

Update segment resource to emit entire pool list section if the values are not
set. Otherwise, the backend will simply fail the request.


spikewang:
Validated on dev, preprod and prod.

kenkkim (approved):

kenkkim (approved):

PR #149, opened 2022-08-12
AK-27089 Enable advanced options for endpoint of connector_ipsec

spikewang:
Enable and complete the advanced_options for each endpoint block in connector_ipsec.

+ Update and complete the advanced_options.
+ Update all documentation related to advanced_options.
+ Update example for advanced_options.

### DIFF
--- a/alkira/resource_alkira_connector_ipsec.go
+++ b/alkira/resource_alkira_connector_ipsec.go
@@ -2,6 +2,7 @@ package alkira
 
 import (
 	"log"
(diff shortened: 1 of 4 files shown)


spikewang:
@jyothsna48 This should be it. I have tested various cases, and you could verify. One interesting finding is that backend actually updated `time` fields automatically in certain cases.

kenkkim (approved):

mac-alkira (approved):

PR #150, opened 2022-08-15
AK-26716 Fixed PAN version bug / made zones_to_groups field optional

mac-alkira:


alkira/resource_alkira_service_pan.go
```diff
@@ -454,6 +454,7 @@ func generateServicePanRequest(d *schema.ResourceData, m interface{}) (*alkira.S
 
 	service := &alkira.ServicePan{
 		BillingTagIds:               convertTypeListToIntList(d.Get("billing_tag_ids").([]interface{})),
+		Bundle:                      d.Get("bundle").(string),
```
spikewang, line 457: Thanks! This is probably missing from the READ function as well? Could we fix that.

spikewang (approved):

PR #151, opened 2022-08-16
AK-27231 Create acceptance.yml

SpaSmALX:
Starting workflow acceptance test file


.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,51 @@
+name: Acceptance Test CI
```
spikewang, line 1: "Acceptance Test" should be good enough. The word will show up in the Action page. Action is already CI.

spikewang (approved):
Looks good

spikewang:
Looks good! Let's go!

PR #155, opened 2022-08-22
AK-27442 Fix credential id included in PUT requests for infoblox instances

mac-alkira:


alkira/resource_alkira_service_infoblox.go
```diff
@@ -149,6 +149,11 @@ func resourceAlkiraInfoblox() *schema.Resource {
 							Type:        schema.TypeInt,
 							Computed:    true,
 						},
+						"credential_id": {
```
spikewang, line 152: I thought we want to hide the instance credential like other services?

spikewang: my bad, I see it now...

spikewang (approved):

PR #161, opened 2022-08-23
AK-27508 Add empty string validation in connector_ipsec

spikewang:
Add empty string validations for preshared_keys and billing_tag_ids in
connector_ipsec.

### DIFF
--- a/alkira/helper.go
+++ b/alkira/helper.go
@@ -47,6 +47,8 @@ func convertTypeListToIntList(in []interface{}) []int {
 }
 
 func convertTypeListToStringList(in []interface{}) []string {
+	log.Printf("[DEBUG] Convert TypeList %v", in)
+
 	if in == nil || len(in) == 0 {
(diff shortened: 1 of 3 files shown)


spikewang:
empty string or null will fail the check:

```
╷
│ Error: expected "endpoint.1.preshared_keys.0" to not be an empty string or whitespace
│
│   with alkira_connector_ipsec.test,
│   on main.tf line 90, in resource "alkira_connector_ipsec" "test":
│   90:       preshared_keys       = ["  ", ""]
│
╵
╷
│ Error: expected "endpoint.1.preshared_keys.1" to not be an empty string or whitespace
│
│   with alkira_connector_ipsec.test,
│   on main.tf line 90, in resource "alkira_connector_ipsec" "test":
│   90:       preshared_keys       = ["  ", ""]
│
```

This is probably the best we could do to avoid state diff... for now

kenkkim (approved):

PR #177, opened 2022-09-08
AK-27486 Make license_key optional in PAN credential

spikewang:


spikewang:
This may be not necessary anymore once we retired both PAN credentials, but just put it up just in case.

PR #190, opened 2022-09-23
AK-28226 Change management_server_segment to be an int id like PAN

mac-alkira:


alkira/resource_alkira_service_fortinet.go
```diff
@@ -89,9 +91,9 @@ func resourceAlkiraServiceFortinet() *schema.Resource {
 				Optional:    true,
 			},
 			"management_server_segment": {
-				Description: "The segment used to access the management server. This segment " +
+				Description: "The segment ID used to access the management server. This segment " +
```
spikewang, line 94: Should we call the identifier `management_server_segment_id` to be consistent?

spikewang:
Thanks @mac-alkira! Should we update example as well?

PR #194, opened 2022-10-07
AK-28599 multiple zones per segment

mac-alkira:


alkira/helper.go
```diff
@@ -52,10 +52,14 @@ func expandSegmentOptions(in *schema.Set, m interface{}) (alkira.SegmentNameToZo
 			return nil, errors.New("segment_option fields cannot be nil")
```
spikewang, line 52: I am not sure if this is true... e.g. for service_pan, I think the zone to group mapping (now it's SegmentOptions) is optional. Not true anymore?

spikewang (approved):

alkira/helper.go
mac-alkira: I'll double check the docs and run a test to see. If not required will remove.

PR #215, opened 2022-10-26
AK-20014 Add credential ssh key pair resource

spikewang:
Add new resource to store SSH key pair and allow it to be used in connector_cisco_sdwan.


spikewang:
example was added in the test repo...

kenkkim (approved):

PR #244, opened 2022-12-12
AK-30652 Toggle default advertise_internet_exit of policy_routing to …

spikewang:
…true

Toggle default advertise_internet_exit of policy_routing to true.


Alshadex (approved):

kenkkim (approved):

spikewang:
The backend still has validation that INBOUND policy can't set the option... No need to make this change for now.

PR #252, opened 2023-01-11
AK-31225 Update advertise_internet_exit of routing_policy with defaul…

spikewang:
…t value

Update advertise_internet_exit of routing_policy with default value true, but the field should be only applicable when policy is OUTBOUND. For INBOUND policy, the field should be NULL in the request.


spikewang:
This is a quick fix from the TF side to unblock the problem. I would expect the API should inherit the same behaviour instead of expecting NULL field when it's INBOUND policy.

kenkkim (approved):

PR #257, opened 2023-02-07
AK-30862 support for gcp subnet selection

snehachandrak:


alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -55,6 +55,26 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"vpc_subnet": {
+				Description: "The list of subnets of the target GCP VPC for routing purpose." +
```
spikewang, line 59: For multi lines break, you may want to add an extra space at last like `...purpose. "`. Otherwise, the space won't show up once generating docs.

alkira/resource_alkira_connector_gcp_vpc_helper.go
```diff
@@ -1,39 +1,74 @@
...
+			r := alkira.UserInputPrefixes{}
+			t := subnet.(map[string]interface{})
+			if v, ok := t["id"].(string); ok {
+				r.Id = v
+			}
+			if v, ok := t["cidr"].(string); ok {
+				r.Value = v
+			}
+
+			r.Type = "SUBNET"
```
spikewang, line 64: Both Id and Value is optional, what if both are empty? Type will be still set to `SUBNET`... so some half empty struct will be sent to backend... which failed here many times before...

docs/resources/connector_gcp_vpc.md
```diff
@@ -10,8 +10,16 @@ description: |-
 
 Manage GCP Cloud Connector.
 
+## Routing Options
```
spikewang, line 13: You don't edit this file directly. It will be generated by `make doc` from templates and source codes... You could add a new template in `templates/resources/` as `connector_gcp_vpc.md.tmpl` similar to others in the directory.. and add those descriptions in the template.

docs/resources/connector_gcp_vpc.md
```diff
@@ -26,6 +34,66 @@ resource "alkira_connector_gcp_vpc" "gcp_vpc1" {
 }
 ```
 
+You could adjust routing options on the VPC by using
+`gcp_routing` and `prefix_list_ids` along with resource
+`alkira_policy_prefix_list`:
+
+```terraform
```
spikewang, line 41: For examples, you will need to add new one in `examples` and add multi examples in the template file. Refer to connector-aws-vpc.

spikewang (approved):

spikewang (changes requested):
Please fix the documentation part and create a new template and re-generate docs

alkira/resource_alkira_connector_gcp_vpc_helper.go
snehachandrak: added empty check in line 64 and tested.

spikewang (approved):

PR #258, opened 2023-02-07
AK-31170 update docs for service chaining support on rule

snehachandrak:


alkira/resource_alkira_policy_rule.go
```diff
@@ -105,13 +105,15 @@ func resourceAlkiraPolicyRule() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"ALLOW", "DROP"}, false),
 			},
 			"rule_action_service_types": {
-				Description: "Based on the service type, traffic is routed to service of the given type.",
+				Description: "Based on the service type, traffic is routed to service of the given type." +
```
spikewang, line 108: Add a space at the end like this `type. "`, so the generated doc will have a space.

alkira/resource_alkira_policy_rule.go
```diff
@@ -105,13 +105,15 @@ func resourceAlkiraPolicyRule() *schema.Resource {
...
-				Description: "Based on the service type, traffic is routed to service of the given type.",
+				Description: "Based on the service type, traffic is routed to service of the given type." + 
+					"For service chaining, both PAN and ZIA service types can be selected but must follow order.",
 				Type:        schema.TypeList,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Optional:    true,
 			},
 			"rule_action_service_ids": {
-				Description: "Based on the service IDs, traffic is routed to the specified services.",
+				Description: "Based on the service IDs, traffic is routed to the specified services." +
```
spikewang, line 115: Add one extra space

alkira/resource_alkira_policy_rule.go
```diff
@@ -105,13 +105,15 @@ func resourceAlkiraPolicyRule() *schema.Resource {
...
+					"For service chaining, both PAN and ZIA service ids can be selected but PAN serviceIds must be follow by Zscaler serviceIds.",
```
spikewang, line 116: We don't have a thing called service ids... It's better to refer to the resource directly, like this:

```
For service chaining, both `service_pan` and `service_zscaler`'s IDs could be added here, but ID of `service_pan` must be by followed by ID of `service_zscaler` for now.
```

docs/resources/policy_rule.md
```diff
@@ -29,6 +31,38 @@ resource "alkira_policy_rule" "rule" {
 }
 ```
 
+To configure service chaining, multiple service types can be selected. Service chaining is currently supported only with PAN and Zscaler services.
```
spikewang, line 34: We don't edit those documentation directly. They are generated from source code, templates and examples. If you want to add more examples, you could add it to `examples` directory named after the resource. There should be some examples there.

spikewang (changes requested):
Please update the documentation.

spikewang (approved):

PR #259, opened 2023-02-15
AK-32224 Update alkira-client-go to v1.37.3 with refactored client

spikewang:
Update alkira-client-go to v1.37.3 with the latest refactoring with generics. This also brings bunch of changes for better provisioning support.

+ Enable provisioning option in provider init.
+ Add computed provision_state for most provision-able resources.
+ Use the new generic APIs for all resources.


spikewang:
This one is getting too big and what we have so far should be good enough for working properly. Let's get it in firstly. @Alshadex, your new plugin-framework change could also rebase on this once it's in

Alshadex (approved):

PR #261, opened 2023-02-17
AK-32422 Refactor segment helper functions and usage across resource

spikewang:
Refactor segment helper functions to make it more generic and update 
all resources to use the new helper functions.


spikewang:
The acceptance tests was quickly fixed for now till the hot fix was in.

PR #265, opened 2023-02-24
AK-32496 Add new fq_id field to gcp connector to support gcp subnets …

camjrichards:
…with gcp terraform resources

### DIFF
--- a/alkira/resource_alkira_connector_gcp_vpc.go
+++ b/alkira/resource_alkira_connector_gcp_vpc.go
@@ -75,9 +75,15 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"id": {
-							Description: "The Id of the subnet.",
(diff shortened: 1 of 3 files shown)


alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -75,10 +75,16 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"id": {
-							Description: "The Id of the subnet.",
+							Description: "The numeric Id of the subnet. This is the Id returned by the GCP REST API.",
```
spikewang, line 78: Could we tweak all `Id` to `ID` to be consistent across?

spikewang (approved):

alkira/resource_alkira_connector_gcp_vpc.go
camjrichards: Done

PR #268, opened 2023-03-06
AK-32935 Handle provision state separately from all CRUD operations

spikewang:
Handle provisioning state separately from all CRUD operations.

+ Use diag packages across all resources to have better error handling.
+ Track provisioning failures as a "WARNING" instead of failing and stopping entire TF execution.
+ Update error messages with better information.
+ Update alkira-client-go to v1.37.8 to pick up corresponding changes from client functions.

### DIFF
--- a/alkira/helper.go
+++ b/alkira/helper.go
@@ -37,7 +37,7 @@ func expandSegmentOptions(in *schema.Set, m interface{}) (alkira.SegmentNameToZo
 		}
 
 		if v, ok := optionsCfg["segment_id"].(string); ok {
-			seg, err := segmentApi.GetById(v)
+			seg, _, err := segmentApi.GetById(v)
 
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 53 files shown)


spikewang:
With this new model, we track API errors and provisioning errors separately, since they are totally two separate operations from backend today...

Under this new workflow, provision failures will be considered "partially" failure for a resource. As long as API went through, we will still track the resource from TF state, but a warning will be displayed to indicate that provisioning failed. 

```
╷
│ Warning: PROVISION FAILED
│
│   with alkira_group.tftest1,
│   on main.tf line 16, in resource "alkira_group" "tftest1":
│   16: resource "alkira_group" "tftest1" {
│
│ client-create(client-c3a41f59-777a-4277-ab50-b5f13fa7d2c6): failed to get provision request ID
```

Alshadex (approved):

PR #274, opened 2023-05-02
AK-32666 add support for alkira flow collector resource

amitmandke:


alkira/provider.go
```diff
@@ -83,6 +83,7 @@ func Provider() *schema.Provider {
 			"alkira_service_zscaler":              resourceAlkiraServiceZscaler(),
 			"alkira_service_pan":                  resourceAlkiraServicePan(),
 			"alkira_tenant_network":               resourceAlkiraTenantNetwork(),
+			"alkira_flow_collector":               resourceAlkiraFlowCollector(),
```
spikewang, line 86: This list should be in alphabetical order. Could you move it up?

alkira/resource_alkira_flow_collector.go
```diff
@@ -0,0 +1,250 @@
...
+				Computed:    true,
+			},
+			"collector_type": {
+				Description:  "The type of the flow collector.",
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "GENERIC",
+				ValidateFunc: validation.StringInSlice([]string{"GENERIC"}, false),
+			},
+			"enabled": {
```
spikewang, line 57: I think we usually have a default value for this?

spikewang (approved):
Just couple questions

alkira/resource_alkira_flow_collector.go
amitmandke: this one is user choice as this can cause them to start billing on the flow collector. User can create and keep it disabled and turn it on only when they want to collect flows.

alkira/provider.go
amitmandke: done!

PR #285, opened 2023-06-29
AK-35660 Add support for authentication via API key

camjrichards:


alkira/provider.go
```diff
@@ -126,7 +132,7 @@ func envDefaultFunc(k string) schema.SchemaDefaultFunc {
 }
 
 func alkiraConfigure(d *schema.ResourceData) (interface{}, error) {
-	alkiraClient, err := alkira.NewAlkiraClient(d.Get("portal").(string), d.Get("username").(string), d.Get("password").(string), d.Get("provision").(bool))
+	alkiraClient, err := alkira.NewAlkiraClientAPIKeySupport(d.Get("portal").(string), d.Get("username").(string), d.Get("password").(string), d.Get("api_key").(string), d.Get("provision").(bool))
```
spikewang, line 135: Left another comment in the client-go, I think we don't really need a new function. It's no point...

spikewang (changes requested):
Once you have the alkira-client-go fixed, I will bring a new client over and this could go in.

spikewang (approved):

PR #287, opened 2023-08-08
AK-36293 support optional segment in alkira flow collector resource

amitmandke:


alkira/resource_alkira_flow_collector.go
```diff
@@ -59,6 +59,13 @@ func resourceAlkiraFlowCollector() *schema.Resource {
 				Type:        schema.TypeBool,
 				Required:    true,
 			},
+			"segment": {
```
spikewang, line 62: This is our segment I guess? We decided to always use `segment_id` from Terraform instead of Segment name. It's confusing. Could you change it to `segment_id` and we will need to convert. There are many examples in other resources.

spikewang (changes requested):

alkira/resource_alkira_flow_collector.go
amitmandke: Done!

spikewang (approved):
Looks good to me!

spikewang (approved):

PR #295, opened 2023-10-16
AK-37682 Add allow list id to infoblox service

camjrichards:


alkira/resource_alkira_service_infoblox.go
```diff
@@ -230,6 +230,11 @@ func resourceAlkiraInfoblox() *schema.Resource {
 				Required:     true,
 				ValidateFunc: validation.StringIsNotEmpty,
 			},
+			"allow_list_id": {
```
spikewang, line 233: What's the naming of the field on UI for this? It's called `allow list`? It's confusing, since we never have a resource called `allow list` anywhere.

alkira/resource_alkira_service_infoblox.go
```diff
@@ -420,5 +425,6 @@ func generateInfobloxRequest(d *schema.ResourceData, m interface{}) (*alkira.Ser
 		Name:             name,
 		Segments:         segmentNames,
 		ServiceGroupName: d.Get("service_group_name").(string),
+		AllowListId:      d.Get("allow_list_id").(int),
```
spikewang, line 428: This is only the request, you will also need to add it to READ function.

alkira/resource_alkira_service_infoblox.go
spikewang: hmmmmm, let me confirm with UI folk... since if we call it whitelist_prefix_list, that's also not there...

PR #298, opened 2023-11-06
AK-39332 Add Custom routing options to nat rule resource

camjrichards:


alkira/resource_alkira_policy_nat_rule.go
```diff
@@ -151,7 +151,23 @@ func resourceAlkiraPolicyNatRule() *schema.Resource {
...
+							Elem:        &schema.Schema{Type: schema.TypeString},
+							Optional:    true,
+						},
+						"src_addr_translation_routing_track_prefix_list_ids": {
+							Description: "The list of prefix list IDs.",
+							Type:        schema.TypeList,
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+							Optional:    true,
+						},
+						"src_addr_translation_routing_invalidate_track_prefixes": {
```
spikewang, line 167: If we decided to call those fields with prefix `src_addr_translation_routing_track_`, then let's keep them consistent.

alkira/resource_alkira_policy_nat_rule_helper.go
```diff
@@ -110,6 +110,24 @@ func expandPolicyNatRuleAction(in *schema.Set) *alkira.NatRuleAction {
 			e.IpType = v
 		}
 
+		if v, ok := actionValue["src_addr_translation_routing_track_prefixes"].([]interface{}); ok {
```
spikewang, line 113: I don't see where we set those values in READ function...

spikewang (changes requested):

alkira/resource_alkira_policy_nat_rule_helper.go
camjrichards: In the NAT rule read function it looks like we're not accessing any of the fields for match or action - https://github.com/alkiranet/terraform-provider-alkira/blob/73a65dd3bd9d43b12d7582ad3ce8bb52cb226ebb/alkira/resource_alkira_policy_nat_rule.go#L261

Am I misunderstanding something or should we add all the relevant fields in that function?

spikewang (approved):

PR #299, opened 2023-11-15
AK-39039 Add customer ASN field to Azure vnet and GCP vpc connectors

camjrichards:


alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -172,6 +172,12 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 					"SMALL", "MEDIUM", "LARGE", "2LARGE",
 					"4LARGE", "5LARGE", "10LARGE", "20LARGE"}, false),
 			},
+			"customer_asn": {
+				Description: "A specific BGP ASN for the connector. This cannot be specified for connectors in `VNET_PEERING` mode. This field cannot be updated" +
```
spikewang, line 176: Please refer to the existing field in the resource, It should be like:

This cannot be specified when `connector_mode` is `VNET_PEERING`.

alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -172,6 +172,12 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
...
+					"once the connector has been provisioned. The ASN can be any private ASN (64512 - 65534, 4200000000 - 4294967294) that is not used elsewhere in the network.",
```
spikewang, line 177: For here, it will be good to keep values in the code block of Markdown:

The value of the field could be any private ASN (`54512 - 65534`,  `4200000000 - 4294967294`)

alkira/resource_alkira_connector_gcp_vpc.go
spikewang: Also, the comment above is for azure-vent. Please fix this for GCP VPC. There is no VNET in GCP.

spikewang (changes requested):
Please fix the GCP-VPC description.

camjrichards:
> Please fix the GCP-VPC description.

Have updated connector descriptions

spikewang (approved):

PR #303, opened 2023-12-22
AK-40766 - SNAT supports only DYANAMIC_IP_AND_PORT

pratikjoshi11:


spikewang (approved):
Please fix the API doc too

PR #305, opened 2024-02-01
AK-41730 Add resource connector_remote_access

spikewang:
Add new resource connector_remote_access.


spikewang:
tested the basic one on production. Will need further testings for some other use cases...

kenkkim (approved):

PR #309, opened 2024-05-21
AK-44577 set udrLists on vnet routing only if set in request instead …

snehachandrak:
…of initializing with null arrays

### DIFF
--- a/alkira/resource_alkira_connector_azure_vnet_helper.go
+++ b/alkira/resource_alkira_connector_azure_vnet_helper.go
@@ -221,7 +221,10 @@ func constructVnetRouting(d *schema.ResourceData) (*alkira.ConnectorVnetRouting,
 		ExportOptions: exportOptions,
 		ImportOptions: importOptions,
 		ServiceRoutes: serviceRoutes,
-		UdrLists:      udrLists,
+	}
+
+	if len(udrLists.Subnets) > 0 || len(udrLists.Cidrs) > 0 {
+		vnetRouting.UdrLists = udrLists
 	}
 
 	return &vnetRouting, nil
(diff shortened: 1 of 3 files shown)


alkira/resource_alkira_connector_azure_vnet_helper.go
```diff
@@ -221,7 +221,10 @@ func constructVnetRouting(d *schema.ResourceData) (*alkira.ConnectorVnetRouting,
 		ExportOptions: exportOptions,
 		ImportOptions: importOptions,
 		ServiceRoutes: serviceRoutes,
-		UdrLists:      udrLists,
+	}
+
+	if len(udrLists.Subnets) > 0 || len(udrLists.Cidrs) > 0 {
```
moiz-alkira, line 226: Can udrLists be nil?

moiz-alkira (commented):
Any unit tests for this?

alkira/resource_alkira_connector_azure_vnet_helper.go
snehachandrak: no, it is being initialized in line 87

spikewang: If this happens to UdrList, same thing will happen to ServiceRoutes, nope? If we choose to use this kind super nested arrays, this will be just impossible to maintain.

spikewang (approved):
To be honest, doesn't really like this hack, null array should be handled or ignored.

snehachandrak:
> To be honest, doesn't really like this hack, null array should be handled or ignored.

We are working on another fix to ignore json nodes with null entries. This is to unblock while we get that commit in and it needs some soak time.

PR #310, opened 2024-05-23
AK-43858 Add new optional overlappingRouteValidationEnabled to segmen…

spikewang:
…t resource

Add new optional overlappingRouteValidationEnabled to segment resource.


spikewang:
will need to have the new alkira-client-go published firstly then merge this one. Tested on PREPROD and PROD without feature enabled to ensure backward compatibility.

PR #313, opened 2024-06-06
AK-22097 Ath on AWS CXPs

asim-ak:
added peering_gateway_cxp resource and data source.
updated azure_vnet_connector to support cxpPeeringGatewayId


alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -64,7 +64,8 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 				Optional: true,
 				Default:  "VNET_GATEWAY",
 				ValidateFunc: validation.StringInSlice([]string{
-					"VNET_GATEWAY", "VNET_PEERING"}, false),
+					"VNET_GATEWAY", "VNET_PEERING",
+				}, false),
```
spikewang, line 68: what's changed here except new line?

alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -109,7 +110,8 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 				Default:  "ADVERTISE_DEFAULT_ROUTE",
 				ValidateFunc: validation.StringInSlice([]string{
 					"ADVERTISE_DEFAULT_ROUTE",
-					"ADVERTISE_CUSTOM_PREFIX"}, false),
+					"ADVERTISE_CUSTOM_PREFIX",
+				}, false),
```
spikewang, line 114: same here, this is not really related to what this PR should do?

alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -243,18 +248,21 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"cxp_peering_gateway_id": {
```
spikewang, line 251: I think this field should be consistent with resource name, which is `peering_gateway_cxp_id`

alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -403,12 +406,10 @@ func resourceConnectorAzureVnetDelete(ctx context.Context, d *schema.ResourceDat
 
 // generateConnectorAzureVnetRequest generate request for connector-azure-vnet
 func generateConnectorAzureVnetRequest(d *schema.ResourceData, m interface{}) (*alkira.ConnectorAzureVnet, error) {
-
 	//
 	// Segment
 	//
 	segmentName, err := getSegmentNameById(d.Get("segment_id").(string), m)
-
```
spikewang, line 411: not sure why all those lines change, seems like not related. Please keep the change related to this PR, so it's easier to see what's the real change.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+		CreateContext: resourceAlkiraPeeringGatewayCxpCreate,
+		ReadContext:   resourceAlkiraPeeringGatewayCxpRead,
+		UpdateContext: resourceAlkiraPeeringGatewayCxpUpdate,
+		DeleteContext: resourceAlkiraPeeringGatewayCxpDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "The name of the Peering Gateway.\nOnce deployed this property can not be changed. All the changes will be ignored.",
```
spikewang, line 26: Try to keep all those description at 80 per line... so it will be much easier to manage.

spikewang: And "all changes will be ignored" is not a good behaviour... we should fix that and API should report errors back if someone tries to update this field.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"description": {
+				Description: "Description of the Peering Gateway.",
```
spikewang, line 31: Make the naming consistent, `Description of the CXP Peering Gateway`, since there is other peering gateway resources.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"cxp": {
+				Description: "The CXP to which the Gateway is attached.\nOnce deployed this property can not be changed. All the changes will be ignored.",
```
spikewang, line 36: same here... we simple can't say that "all changes will be ignored"... API should return proper error...

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+
+			// TODO: change cloud_provider to be a Required value when more cloud providers are added and remove the default value.
+			"cloud_provider": {
+				Description: "The cloud provider on which the gateway will be created.\nOnce deployed this property can not be changed. All the changes will be ignored.",
+				Type:        schema.TypeString,
+				// Required:    true,
```
spikewang, line 45: Remove

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+				Optional: true,
+				Default:  "AZURE",
+			},
+			"cloud_region": {
+				Description: "The cloud region on which the ATH will be created. Eg : eastus , westus.\nOnce deployed this property can not be changed. All the changes will be ignored.",
```
spikewang, line 50: Rework this to be more clear:

The region of the specified cloud provider on which the resource should be created. E.g.  if `cloud_provider` is `AZURE`, the region could be `eastus`.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+			"cxp": {
+				Description: "The CXP to which the Gateway is attached.\nOnce deployed this property can not be changed. All the changes will be ignored.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+
+			// TODO: change cloud_provider to be a Required value when more cloud providers are added and remove the default value.
+			"cloud_provider": {
+				Description: "The cloud provider on which the gateway will be created.\nOnce deployed this property can not be changed. All the changes will be ignored.",
```
spikewang, line 43: I don't think any other providers are supported at this time, so say it clearly in Description:

The cloud provider on which the resource will be created. Default value is `AZURE` and only `AZURE` is supported for now.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,173 @@
...
+				Optional: true,
+				Default:  "AZURE",
+			},
+			"cloud_region": {
+				Description: "The cloud region on which the ATH will be created. Eg : eastus , westus.\nOnce deployed this property can not be changed. All the changes will be ignored.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"segment": {
+				Description: "The name of the segment in which the gateway is created.\nOnce deployed this property can not be changed. All the changes will be ignored.",
```
spikewang, line 55: We do use `segment_id` in TF. This should be consistent with all other resources. We will take `segment_id` and convert to segment name internally. Refer to other resources.

spikewang (changes requested):

examples/resources/alkira_peering_gateway_cxp/resource.tf
```diff
@@ -0,0 +1,7 @@
+resource "alkira_peering_gateway_cxp" "test1" {
+  name         = "tf-test-1"
```
spikewang, line 2: For example, we try to move away from referring to our internal test naming. I am fixing all other examples. so we could just use name `example`. Tweak all `test` to `example`.

alkira/resource_alkira_connector_azure_vnet.go
asim-ak: sorry, this is caused by "gofumpt" formatting.

alkira/resource_alkira_peering_gateway_cxp.go
asim-ak: updated

alkira/resource_alkira_peering_gateway_cxp.go
asim-ak: We had a discussion during the ACG meeting, basically if a resource cannot be updated it would just ignore the changes, if the update was conditional , it would throw an error. 
So in this case the changes would be just ignored. I was asked to add this note in the documentation.

alkira/resource_alkira_peering_gateway_cxp.go
asim-ak: done.

alkira/resource_alkira_peering_gateway_cxp.go
asim-ak: done.

alkira/resource_alkira_peering_gateway_cxp.go
asim-ak: done.

examples/resources/alkira_peering_gateway_cxp/resource.tf
asim-ak: done.

alkira/resource_alkira_peering_gateway_cxp.go
asim-ak: done.

alkira/resource_alkira_connector_azure_vnet.go
spikewang: lol, we used the standard `gofmt`, so it should be okey. It's funny that the tool says that it won't generate diff from `gofmt`. Hehe.

alkira/resource_alkira_peering_gateway_cxp.go
spikewang: I disagree. This doesn't belong to ACG team. It's just a bad user experience. If backend doesn't want to fix this, please do validation in `update` function.

spikewang: Even we wanted to have this description. It should be like this and have line break at 80 like this:

```
Description: "The CXP to which the Gateway is attached to. " +
                      "Once deployed, this field can't be updated.",
```

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,180 @@
...
+	d.Set("state", resource.State)
+
+	return nil
+}
+
+func resourceAlkiraPeeringGatewayCxpUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	// INIT
+	api := alkira.NewPeeringGatewayCxp(m.(*alkira.AlkiraClient))
+
+	request, err := generateAlkiraCxpPeeringGatewayRequest(d, m)
```
spikewang, line 137: If we don't actually support update, then let's error out here directly, instead of commenting on almost every single field, which doesn't make any sense.

spikewang (changes requested):

alkira/resource_alkira_connector_azure_vnet.go
spikewang: Again, I am not really sure why benefit to have this new format. `gofmt` is built-in and generic enough across all our Golang services. We should stick with that.

spikewang (changes requested):

asim-ak:
@spikewang regarding 'gofumpt' my understanding is that it is much more stricter version of 'gofmt' , which leads it to be more consistent. 
I've reverted to to 'gofmt' for now.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+				Optional:    true,
+			},
+			"cxp": {
+				Description: "The CXP to which the Gateway is attached." +
+					"Once deployed, this property can not be changed.",
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			// TODO: change cloud_provider to be a Required value when more cloud providers are added and remove the default value.
```
spikewang, line 42: Remove this. This is not needed. We will have task tracking this if new provider is added later on.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+			"cloud_provider": {
+				Description: "The cloud provider on which the resource will be created." +
+					"Default value is AZURE and only AZURE is supported for now.",
```
spikewang, line 45: Add AZURE into markdown code by using "\`". It should be like \`azure\`.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "AZURE",
+			},
+			"cloud_region": {
+				Description: "The region of the specified cloud provider on which the resource should be created." +
+					"E.g. if cloud_provider is AZURE, the region could be eastus.",
```
spikewang, line 52: same here, cloud_provider should be in as code block. The line seems like still too long to me. 80 break.

"The region of the specified ``cloud_provider`` on which the resource " +
"should be created. For example, if ``cloud_provider`` is ``AZURE``, the " +
"region could be ``eastus``.",

Also, notice that at the line end, you will need to leave a space there.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"segment_id": {
+				Description: "The ID of the segment in which the gateway is created." +
+					"Once deployed, this property can not be changed.",
```
spikewang, line 58: Remove this property can not be changed.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+		ReadContext:   resourceAlkiraPeeringGatewayCxpRead,
+		UpdateContext: resourceAlkiraPeeringGatewayCxpUpdate,
+		DeleteContext: resourceAlkiraPeeringGatewayCxpDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "The name of the Peering Gateway." +
+					"Once deployed, this property can not be changed.",
```
spikewang, line 26: Remove this line. No need.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+				Required: true,
+			},
+			"description": {
+				Description: "Description of the CXP Peering Gateway.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"cxp": {
+				Description: "The CXP to which the Gateway is attached." +
+					"Once deployed, this property can not be changed.",
```
spikewang, line 37: Remove this line.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,193 @@
...
+	d.Set("cloud_region", resource.CloudRegion)
+	d.Set("segment_id", segmentId)
+	d.Set("state", resource.State)
+
+	return nil
+}
+
+func resourceAlkiraPeeringGatewayCxpUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	// checking if only description field is changed
+	if d.HasChange("description") && !d.HasChanges("name", "cloud_provider", "cloud_region", "segment_id") {
```
spikewang, line 139: Honestly, if the only field could be changed is description, we should just fail and return. It's pointless to just check description. Just return directly. Simple and clean.

spikewang (changes requested):
I will show you how to use the template file.

alkira/resource_alkira_peering_gateway_cxp.go
```diff
@@ -0,0 +1,194 @@
...
+	api := alkira.NewPeeringGatewayCxp(m.(*alkira.AlkiraClient))
+
+	request, err := generateAlkiraCxpPeeringGatewayRequest(d, m)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// UPDATE
+	_, err, _ = api.Update(d.Id(), request)
+
```
spikewang, line 158: We need to check this `err` before return or just simplify this update. We don't need to support update at all. Just disable update return directly.

spikewang (changes requested):

spikewang (approved):

PR #315, opened 2024-06-19
AK-45580 Add new resource for virtual_manager_azure

asim-ak:
added `alkira_virtual_network_manager_azure` resource to manage AVNMs.
updated `resource_alkira_connector_azure_vnet` to support `direct_inter_vpc_communication_group`.
updated `resource_alkira_group_direct_inter_connector` to support Azure direct inter connect groups.
added examples and docs.


alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -247,6 +247,11 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 				Type:        schema.TypeInt,
 				Optional:    true,
 			},
+			"direct_inter_vnet_communication_group": {
+				Description: "Direct inter-vnet communication group.",
```
spikewang, line 251: I think we need to explain this a bit, since it's referring to the resource `alkira_group_direct_inter_connector`. The group name is generic. It's not specific to just inter VNET. Do we plan to have like inter VPC-VNET?

alkira/resource_alkira_group_direct_inter_connector.go
```diff
@@ -51,23 +52,29 @@ func resourceAlkiraDirectInterConnectorGroup() *schema.Resource {
...
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"AWS_VPC", "AZURE_VNET"}, false),
 			},
 			"provision_state": {
 				Description: "The provisioning state of the resource.",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
+			"virtual_network_manager_azure_id": {
+				Description: "The Azure Virtual Network Manager's Alkira ID.",
```
spikewang, line 74: The description is mis-leading. I think we should just say:

"ID of `virtual_network_manager_azure`.

alkira/resource_alkira_group_direct_inter_connector.go
```diff
@@ -216,3 +207,21 @@ func resourceDirectInterConnectorGroupDelete(ctx context.Context, d *schema.Reso
 
 	return nil
 }
+
+func generateDirectInterConnectorGroupRequest(d *schema.ResourceData, m interface{}) (*alkira.InterConnectorCommunicationGroup, error) {
+
+	// Segment
+	segmentName, _ := getSegmentNameById(d.Get("segment_id").(string), m)
+
```
spikewang, line 215: If we failed to get segmentName, we need to error out.

alkira/resource_alkira_group_direct_inter_connector_helper.go
```diff
@@ -0,0 +1,33 @@
+package alkira
+
+import (
+	"github.com/alkiranet/alkira-client-go/alkira"
+)
+
+// getDirectInterConnectorGroupNameByID gets the direct interconnector group name by its ID.
+
+func getDirectInterConnectorGroupNameByID(id string, m interface{}) (string, error) {
```
spikewang, line 9: Why do we need this? Is it used anywhere? I don't see any references...

alkira/resource_alkira_virtual_network_manager_azure.go
```diff
@@ -0,0 +1,207 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"description": {
+				Description: "The description of the Alkira Azure Virtual Network Manager.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"credential_id": {
+				Description: "ID of credential managed by Credential Manager.",
```
spikewang, line 70: This should refer to the credential resource for azure. We don't have credential manager.

alkira/resource_alkira_virtual_network_manager_azure.go
```diff
@@ -0,0 +1,207 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"subscriptions_in_scope": {
```
spikewang, line 74: This is a bit confusing too... We asked for subscription_id, what subscriptions in scope means...

alkira/resource_alkira_virtual_network_manager_azure.go
```diff
@@ -0,0 +1,207 @@
...
+		state = resource.State
+		time.Sleep(5 * time.Second)
+
+	}
+	return resourceVirtualNetworkManagerAzureRead(ctx, d, m)
+}
+
+func resourceVirtualNetworkManagerAzureUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	// CHECK FOR CHANGES IN ANY FIELD EXCEPT DESCRIPTION AND RETURN ERR.
+	if d.HasChange("description") {
```
spikewang, line 126: If only description could be updated for a resource, I think we should consider this resource not upgradable at all.

spikewang (changes requested):

alkira/resource_alkira_connector_azure_vnet.go
asim-ak: I don't think direct VPC-VNET would be possible since they are on different cloud and all this feature does is allow two VPCs or VNET in the same cloud to use the cloud backbone.

alkira/resource_alkira_group_direct_inter_connector.go
asim-ak: Basically we create a Virtual Network Manager on Azure and they have their own ID, that's why I explicitly mentioned Alkira ID.

alkira/resource_alkira_virtual_network_manager_azure.go
asim-ak: This is same as the peering gateway, we can only update the description. Otherwise we throw an error saying it is an "Invalid Update".

alkira/resource_alkira_virtual_network_manager_azure.go
asim-ak: The subscription id is for the subscription that the AVNM will be created in and the `subscriptions_in_scope` are the subscriptions managed by the AVNM.

asim-ak: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/network_manager

alkira/resource_alkira_virtual_network_manager_azure.go
asim-ak: image
image

This was the description used for `credential_id` in other resources so I tried to keep it consistent

spikewang:
We will close this for now till future use case.

PR #316, opened 2024-06-22
Cleanup udr in azure vnet connector.

asim-ak:
Revert changes to azure vnet connector UDR.
The backend now handles the nil values.
Also this code did not do anything as the UdrLists struct would be initalized with nil (default values) values when nothing was passed by the helper function.


spikewang (approved):
Hehe thanks

spikewang:
@asim-ak remember we always need a ticket number for any commit. I will re-use the last ticket for this.

PR #318, opened 2024-06-27
Ak 45580 vnet vnet direct interconnect

asim-ak:


spikewang (approved):

sankeerthmogili:
Shouldn't we revert this PR? this change without azure virtual network manager resource doesn't make sense. @asim-ak @spikewang

spikewang:
@sankeerthmogili this is fine. People could still use UI or API to create the resource and pass it in. That’s what QA will do

PR #321, opened 2024-07-04
AK-46114

asim-ak:
updated azure express route to support `SMALL` and `MEDIUM` size.


spikewang (approved):

spikewang:
@asim-ak I think now it's time to remove all those validations from provider... It will be hard to keep up. Let's remove all of them. API should validate the size anyways...

PR #323, opened 2024-07-10
AK-46247 Correct the mac_gateway_address in connector_azure_expressroute

asim-ak:
fixed the field name


spikewang:
This will need more fixes from the provider itself.

spikewang (approved):

PR #329, opened 2024-07-25
AK-46671 - Sync connector sizes with terraform client

asim-ak:
Sync connector sizes with terraform client

### DIFF
--- a/alkira/resource_alkira_connector_azure_vnet.go
+++ b/alkira/resource_alkira_connector_azure_vnet.go
@@ -227,12 +227,12 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 			},
 			"size": {
 				Description: "The size of the connector, one of `SMALL`, `MEDIUM`, " +
-					"`LARGE`, `2LARGE`, `4LARGE`, `5LARGE`, `10LARGE`, `20LARGE`.",
+					"`LARGE`, `2LARGE`, `5LARGE`.",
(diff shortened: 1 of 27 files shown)


alkira/resource_alkira_connector_azure_vnet.go
```diff
@@ -227,12 +227,12 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 			},
 			"size": {
 				Description: "The size of the connector, one of `SMALL`, `MEDIUM`, " +
-					"`LARGE`, `2LARGE`, `4LARGE`, `5LARGE`, `10LARGE`, `20LARGE`.",
+					"`LARGE`, `2LARGE`, `5LARGE`.",
 				Type:     schema.TypeString,
 				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					"SMALL", "MEDIUM", "LARGE", `2LARGE`,
-					`4LARGE`, `5LARGE`, `10LARGE`, `20LARGE`}, false),
```
spikewang, line 235: We could actually remove all those validation functions, size is always changing and API should have handled validation

spikewang (approved):

alkira/resource_alkira_connector_azure_vnet.go
asim-ak: I think we have validations for size in the backend, I'll confirm and we can start removing those validations moving forward.

PR #333, opened 2024-08-08
AK-46654 - fix discrepancies between BYOIP resource and data source naming.

asim-ak:
The resource name was `alkira_byoip_prefix` and the data source was named `alkira_byoip`.
Deprecate `alkira_byoip` data source and add `alkira_byoip_prefix` data source


alkira/data_source_alkira_byoip.go
```diff
@@ -1,7 +1,6 @@
 package alkira
 
 import (
-	"github.com/alkiranet/alkira-client-go/alkira"
```
spikewang, line 4: Why this is removed? The old data source will be broken

docs/data-sources/byoip.md
```diff
@@ -6,8 +6,11 @@ description: |-
   Use this data source to get an existing BYOIP Prefix.
 ---
 
+
 # alkira_byoip (Data Source)
 
+~> **NOTE:** This data source is deprecated, use `alkira_byoip_prefix` instead.
```
spikewang, line 12: You can't modify docs directly. It's generated...

spikewang (approved):

PR #334, opened 2024-08-19
AK-47328 - cloud_provider wasn't being sent in the payload

asim-ak:
missed setting the cloud_provider in the payload.


spikewang (approved):
Plese test out next time to avoid this kinds of problems

PR #340, opened 2024-09-15
AK-42866 - Add F5 Load Balancer and F5 vServer.

asim-ak:


alkira/resource_alkira_list_global_cidr.go
```diff
@@ -62,10 +62,11 @@ func resourceAlkiraListGlobalCidr() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"tags": {
-				Description: "A list of associated service types.",
-				Type:        schema.TypeList,
-				Optional:    true,
-				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "Tag of associated service types." +
+					"Can be one of `INFOBLOX`, `CHKPFW`, `CISCO_FTDV_FW` or `F5LB`.",
+				Type:     schema.TypeList,
```
spikewang, line 67: This could be a TypeSet, so it won't have an order

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -0,0 +1,362 @@
...
+				Description: "The segment options as used by your F5 Load Balancer.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"segment_id": {
+							Description: "ID of the segment.",
+							Type:        schema.TypeString,
+							Required:    true,
+						},
+						"nat_pool_prefix_length": {
+							Description: "Prefix length of subnets for the segment.",
```
spikewang, line 97: Do we have a default value for this? It feels like an value with a default

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -0,0 +1,362 @@
...
+							Type:        schema.TypeInt,
+							Optional:    true,
+						},
+						"elb_nic_count": {
+							Description: "Number of nics to allocate for the segment.",
```
spikewang, line 102: nics -> NICs

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -0,0 +1,362 @@
...
+							Type:        schema.TypeInt,
+							Required:    true,
+						},
+					},
+				},
+			},
+			"service_group_name": {
+				Description: "Name of the service group to be associated with the service.",
```
spikewang, line 110: I am not sure if we have a `service_group` concept. It's just group on portal? All those groups are actually just `group`. Could you double-check this one?

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -0,0 +1,362 @@
...
+							Computed: true,
+						},
+						"f5_registration_key": {
+							Description: "Registration key for the F5 load balancer." +
+								" Only required if `license_type` is `BRING_YOUR_OWN`." +
+								" This can also be set by `ALKIRA_F5_REGISTRATION_KEY`" +
+								" environment variable.",
+							Type:        schema.TypeString,
+							Optional:    true,
+							Sensitive:   true,
```
spikewang, line 156: Good idea! Thanks!

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -0,0 +1,362 @@
...
+							Description: "ID of the F5 load balancer credential." +
+								" If the `credential_id` is not passed," +
+								" `f5_username` and `f5_password` is required" +
+								" to create new credentials.",
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+						"version": {
+							Description: "The version of the F5 load balancer instance.",
```
spikewang, line 188: I think instance is our concept. The version will be just F5 Load Balancer.

alkira/resource_alkira_service_f5_lb.go
asim-ak: From Cam : "Service groups are shown separately from user configured groups on the portal, but they are all fundamentally the same."

alkira/resource_alkira_service_f5_lb.go
asim-ak: nat pool prefix length has been removed. so I've removed it from TF as well.

alkira/resource_alkira_service_f5_lb.go
spikewang: so we don't have a `service_group`  resource in TF, it's just group. I would assume that it's the same group? Then why just call this `group`. Then it could be created by `alkira_group`? Does `service_group` have anything special?

alkira/resource_alkira_service_f5_lb_helper.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+		if name, ok := tfInstance["name"]; ok {
+
+			instanceStruct.Name = name.(string)
+		}
+		if licenseType, ok := tfInstance["license_type"]; ok {
+			instanceStruct.LicenseType = licenseType.(string)
+			// if the license type is BRING_YOUR_OWN, we need a registration credential.
+			if licenseType == "BRING_YOUR_OWN" {
+				if regCredId, ok := tfInstance["registration_credential_id"].(string); ok {
```
spikewang, line 39: if you use `tfInstance["registration_credential_id"]` with `(string)`, it may crash if the former one is `null`, so it's better to cast it inside the `if`.

alkira/resource_alkira_service_f5_lb_helper.go
```diff
@@ -0,0 +1,258 @@
...
+						}
+						instanceStruct.RegistrationCredentialId = credentialId
+					} else {
+						instanceStruct.RegistrationCredentialId = regCredId
+					}
+				}
+			}
+		}
+		if version, ok := tfInstance["version"]; ok {
+
```
spikewang, line 64: extra space?

alkira/resource_alkira_service_f5_lb_helper.go
```diff
@@ -0,0 +1,258 @@
...
+		if deploymentType, ok := tfInstance["deployment_type"]; ok {
+			instanceDeployment.Type = deploymentType.(string)
+		}
+		if deploymentOption, ok := tfInstance["deployment_option"]; ok {
+			instanceDeployment.Option = deploymentOption.(string)
+		}
+		instanceStruct.Deployment = instanceDeployment
+
+		if credId, ok := tfInstance["credential_id"].(string); ok {
```
spikewang, line 79: same here, using `.(string)` may crash if the variable is `nil`.

docs/resources/service_f5_vserver_endpoint.md
```diff
@@ -0,0 +1,47 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "alkira_service_f5_vserver_endpoint Resource - terraform-provider-alkira"
+subcategory: ""
+description: |-
+  
+---
+
+# alkira_service_f5_vserver_endpoint (Resource)
```
spikewang, line 9: How exactly `f5_lib` resource will work with `f5_vserver_endpoint` here?  What's the relationship between those two resources? Any more details? From the below example, `f5_vserver_endpoint` will depend on `f5_lb`.....  and those references of instances are interesting.

spikewang: so what will happen if we provision? I would guess both resources have been provisioned together otherwise, it will be timing issue? Did we try to provision?

asim-ak: each endpoint is linked to an instance/s of the F5 service. 
Creation worked fine, manual provision worked fine, vserver `provision=true` had some issue, we have a fix. I'm currently testing it out. Will update here once it's done.

alkira/resource_alkira_service_f5_lb.go
spikewang: if it's created automatically, why it's a required field here...

docs/resources/service_f5_vserver_endpoint.md
spikewang: 👍

alkira/resource_alkira_service_f5_lb.go
asim-ak: I actually don't know , I'm assuming just for naming or reference ?

AK-Benjamin: We also use service_group_name in INFOBLOX and PAN.
https://github.com/alkiranet/terraform-provider-alkira/blob/896ecb663c206276b4823e1bfbaf198fa3e712f1/alkira/resource_alkira_service_pan.go#L138
https://github.com/alkiranet/terraform-provider-alkira/blob/896ecb663c206276b4823e1bfbaf198fa3e712f1/alkira/resource_alkira_service_infoblox.go#L223
While the group is created automatically, we still want the name to be provided by the customer

spikewang: oh, well @AK-Benjamin, for those two, same issue as well. Our group name needs to be unique globally, so meaning, service group name needs to be unique? How it's supposed to work. The field was there, but it doesn't mean that it's all good. It's always confusing. I would tweak them to be just "service's group".

spikewang:
let's get this to dev to further soak it.

PR #345, opened 2024-11-29
AK-50656 - Update GCP Interconnect Schema.

asim-ak:


alkira/resource_alkira_connector_gcp_interconnect_helper.go
```diff
@@ -29,84 +28,49 @@ func expandGcpInterconnectCustomerGateways(cg []interface{}) ([]alkira.Connector
 	return customerGateways, nil
 }
 
-func expandGcpInterconnectSegmentOptions(so []interface{}, instanceName string, m interface{}) ([]alkira.ConnectorGcpInterconnectSegmentOption, error) {
-	if so == nil || len(so) == 0 {
+func expandGcpInterconnectSegmentOptions(sO []interface{}, m interface{}) ([]alkira.ConnectorGcpInterconnectSegmentOption, error) {
```
snehachandrak, line 31: so instead of s0

asim-ak: I changed it form so to sO because I wanted it to be camel case. `segmentOptions` shortened to `sO` as it is customary to use camelCase and shorthands in golang.

spikewang: could we actually have a meaningful name for this var? `m` and `d` are commonly used in TF plugin code. `sO` or `so` is very confusing.

alkira/resource_alkira_connector_gcp_interconnect_helper.go
```diff
@@ -125,7 +89,7 @@ func expandGcpInterconnectInstances(in []interface{}, so []interface{}, m interf
 		if v, ok := cfgInstance["edge_availability_domain"].(string); ok {
 			newInstance.GcpEdgeAvailabilityDomain = v
 		}
-		if v, ok := cfgInstance["bgp_auth_key_alkira"].(string); ok {
+		if v, ok := cfgInstance["bgp_auth_key"].(string); ok {
```
spikewang, line 92: Is this intentional? I think we have both bgp_auth_key and bgp_auth_key_alkira...

alkira/resource_alkira_connector_gcp_interconnect_helper.go
```diff
@@ -137,71 +101,64 @@ func expandGcpInterconnectInstances(in []interface{}, so []interface{}, m interf
...
 		}
-		newInstance.SegmentOptions = segmentOptions
+
 		instances[i] = newInstance
 	}
 	return instances, nil
 }
 
-func setGcpInterconnectSegmentOptions(d *schema.ResourceData, instance *alkira.ConnectorGcpInterconnectInstance, m interface{}) ([]map[string]interface{}, error) {
+func setGcpInterconnectSegmentOptions(sO []alkira.ConnectorGcpInterconnectSegmentOption, m interface{}) ([]map[string]interface{}, error) {
```
spikewang, line 117: To make this simpler, you could just pass in instance and also fix the variable name.

examples/resources/alkira_connector_gcp_interconnect/resource.tf
```diff
@@ -11,14 +11,13 @@ resource "alkira_connector_gcp_interconnect" "example_gcp_interconnect" {
...
-    advertise_on_prem_routes = true
-    disable_internet_exit    = false
-    customer_gateways {
-      tunnel_count = 2
+    segment_options {
+      segment_id               = alkira_segment.segment1.id
+      advertise_on_prem_routes = true
+      advertise_default_routes = false
+      customer_gateways {
+        tunnel_count = 2
```
spikewang, line 19: This is a bit weird, it says customer_gateways, but the required field is only `tunnel_count`? How many customer_gateways we could actually associated?

spikewang: If it's only 1 block, we could consider to remove this block and change it to:

* customer_gateway_tunnel_count
* customer_gateway_loopback_ip

This way, it makes more sense than a block...

alkira/resource_alkira_connector_gcp_interconnect_helper.go
```diff
@@ -137,71 +101,64 @@ func expandGcpInterconnectInstances(in []interface{}, so []interface{}, m interf
...
 	return segmentOptions, nil
 }
 
 func setGcpInterconnectInstance(d *schema.ResourceData, connector *alkira.ConnectorGcpInterconnect, m interface{}) {
 	var instances []map[string]interface{}
-	var segmentOptions []map[string]interface{}
 	for _, cInstance := range d.Get("instances").([]interface{}) {
 		configInstance := cInstance.(map[string]interface{})
 		for _, aInstance := range connector.Instances {
 			if configInstance["id"].(int) == aInstance.Id ||
```
spikewang, line 154: We don't need this match check anymore. The list should be maintained and stop looping.

spikewang (changes requested):

alkira/resource_alkira_connector_gcp_interconnect_helper.go
asim-ak: Yes , we only have `bgp_auth_key` field in this schema.

examples/resources/alkira_connector_gcp_interconnect/resource.tf
asim-ak: We can have multiple `customer_gateways` .

alkira/resource_alkira_connector_gcp_interconnect.go
```diff
@@ -112,6 +112,55 @@ func resourceAlkiraConnectorGcpInterconnect() *schema.Resource {
...
+										Type:        schema.TypeString,
+										Required:    true,
+									},
+									"advertise_on_prem_routes": {
+										Description: "Advertise on-prem routes. Default is `false`.",
+										Default:     false,
+										Optional:    true,
+										Type:        schema.TypeBool,
+									},
+									"advertise_default_routes": {
```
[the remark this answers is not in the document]
asim-ak: done, the default value for `disableInternetExit` is `false`  in this case, so `advertise_default_route` become `true` by default.

examples/resources/alkira_connector_gcp_interconnect/resource4.tf
```diff
@@ -0,0 +1,26 @@
...
+  loopback_prefixes = ["10.30.0.0/24"]
+  instances {
+    name                     = "instance1"
+    edge_availability_domain = "AVAILABILITY_DOMAIN_1"
+    customer_asn             = 56009
+    bgp_auth_key             = "key"
+    segment_options {
+      segment_id               = alkira_segment.segment1.id
+      advertise_on_prem_routes = true
+      advertise_default_route = false
```
spikewang, line 17: if you run a `terraform fmt` against this file, it will auto-format things neatly.

spikewang (approved):
Should be good for now. Some additional tweaks needed. Let's try this out firstly.

PR #347, opened 2024-12-09
AK-50900 - update setInstance function

asim-ak:


alkira/resource_alkira_service_f5_lb_helper.go
```diff
@@ -163,59 +163,46 @@ func deflateF5SegmentOptions(in alkira.F5SegmentOption, m interface{}) ([]map[st
...
+			"hostname_fqdn":              in.HostNameFqdn,
+			"deployment_option":          in.Deployment.Option,
+			"deployment_type":            in.Deployment.Type,
+		}
 
-			if configInstance["id"].(int) == apiInstance.Id || configInstance["name"].(string) == apiInstance.Name {
-				new = false
-				break
-			}
+		if f5RegistrationKey != "" {
```
spikewang, line 195: At this point, we don't need to check the value of those hidden fields anymore, right? If TF state has not this instance, we could just set it to empty...  so there is no need for check at all. It won't matter anyways.

examples/resources/alkira_service_f5_lb/resource1.tf
```diff
@@ -13,6 +13,16 @@ resource "alkira_service_f5_lb" "example_lb_1" {
     f5_password     = "passwordispassword"
     f5_username     = "admin"
 
+  }
+  instances {
```
spikewang, line 17: I would rename `instances` to `instance`, since each block is 1 instance, but not instances...

alkira/resource_alkira_service_f5_lb_helper.go
```diff
@@ -80,7 +80,7 @@ func expandF5Instances(in []interface{}, m interface{}) ([]alkira.F5Instance, er
 			credId := rawCredId.(string)
 
 			if credId == "" {
-				credentialName := instanceStruct.Name + randomNameSuffix()
+				credentialName := instanceStruct.Name + "credential" + randomNameSuffix()
```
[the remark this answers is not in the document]
asim-ak:
```@NotNull
@Size(max = 256, message = "constraint.size|name length should not be greater than 256")
private String name;```

The limit is 256.

File: alkira/resource_alkira_service_f5_lb_helper.go [hunk 4]
```diff
@@ -80,7 +83,10 @@ func expandF5Instances(in []interface{}, m interface{}) ([]alkira.F5Instance, er
...
+				if len(instanceStruct.Name) > 255 {
+					credentialName = instanceStruct.Name[0:225] + "credential" + randomNameSuffix()
```
spikewang, line 88: lol, it's okey man, worst case, API will just fail... hehe

spikewang (approved):

PR #349, opened 2024-12-14
Update service_f5_vserver_endpoint.md

AK-Swathi:
Added example for 2 instances in vserver endpoint configuration
Removed protocol icmp from examples
Changed port number from -1 to other port numbers


spikewang (commented):
We don't edit those doc files manually. It's all generated. Please do your change in the code to re-generate those changes.

spikewang (changes requested):
Same as the other one, we don't edit file directly. Please sync up with @asim-ak

PR #350, opened 2024-12-16
AK-50900 - update service F5

asim-ak:
`deployment_option` only supports `TWO_BOOT_LOACTION` for now.
`deployment_type` only supports `GOOD` and `LTM_DNS` for now.
update examples and docs based on this.


AK-Benjamin:
We don't allow underscores in names or fqdns.  Take a look at the examples [here](https://github.com/alkiranet/terraform-provider-alkira/tree/task/AK-50900) for reference

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -191,23 +191,22 @@ func resourceAlkiraF5LoadBalancer() *schema.Resource {
 						},
 						"deployment_option": {
 							Description: "The deployment option of the F5 LB instance," +
-								" can be one of `ONE_BOOT_LOCATION` or `TWO_BOOT_LOCATION`." +
+								" Only supports `TWO_BOOT_LOCATION` for now." +
```
spikewang, line 194: If this is the case, could we hide this field till it's becoming optional. Right now, it's truly no option. Just hide it and use `TWO_BOOT_LOCATION` all the time.

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -191,23 +191,22 @@ func resourceAlkiraF5LoadBalancer() *schema.Resource {
...
 							Type:     schema.TypeString,
 							Optional: true,
 						},
 						"deployment_type": {
 							Description: "The deployment type used for the F5 load balancer instance." +
-								" Can be one of `GOOD`, `BETTER`, `BEST`, `LTM_DNS` or `ALL`," +
-								" deployment types `GOOD`, `BETTER` and `BEST`" +
-								" are only applicable to license_type `PAY_AS_YOU_GO`" +
-								" `LTM_DNS` and `ALL` are only applicable to license_type `BRING_YOUR_OWN`",
+								" Can be one of `GOOD` or LTM_DNS`. Deployment type " +
```
spikewang, line 204: Missing ` for LTM_DNS. Let's try to be consistent to use field name, so it's easier to understand:

Deployment type -> deployment_type
license_type -> ``license_type``.

alkira/resource_alkira_service_f5_lb.go
```diff
@@ -189,25 +189,26 @@ func resourceAlkiraF5LoadBalancer() *schema.Resource {
...
-							Description: "The deployment option of the F5 LB instance," +
-								" can be one of `ONE_BOOT_LOCATION` or `TWO_BOOT_LOCATION`." +
-								" Only required when license_type is `BRING_YOUR_OWN`.",
-							ValidateFunc: validation.StringInSlice(
-								[]string{"ONE_BOOT_LOCATION", "TWO_BOOT_LOCATION"},
-								false),
-							Type:     schema.TypeString,
-							Optional: true,
-						},
+						// "deployment_option": {
```
spikewang, line 192: Just remove this for now. When we really need to do this, then bring it back as a new optional field.

alkira/resource_alkira_service_f5_lb_helper.go
```diff
@@ -193,7 +190,6 @@ func setF5Instances(d *schema.ResourceData, ins []alkira.F5Instance) []map[strin
 			"registration_credential_id": in.RegistrationCredentialId,
 			"credential_id":              in.CredentialId,
 			"version":                    in.Version,
-			"deployment_option":          in.Deployment.Option,
```
spikewang, line 196: We should have a preset value for this? Or we could just leave this out? Backend will handle it?

asim-ak: No, TPS has the default value now, so I'm avoiding a hardcoded value here.

spikewang (approved):

PR #351, opened 2025-01-17
AK-51724: Add ilbCredentialId field to internet applications

AK-Benjamin:


vendor/github.com/alkiranet/alkira-client-go/alkira/internet_applications.go
```diff
@@ -39,6 +39,7 @@ type InternetApplication struct {
 	SnatIpv4Ranges                []*InternetApplicationSnatIpv4 `json:"snatIpv4Ranges,omitempty"`
 	Size                          string                         `json:"size"`
 	Targets                       []InternetApplicationTargets   `json:"targets,omitempty"`
+	IlbCredentialId               string                         `json:"ilbCredentialId"`
```
spikewang, line 42: Please make this change in alkira-client-go. This is not the proper way to do it, just to make it easier for testing.

alkira/resource_alkira_internet_applications.go
```diff
@@ -162,6 +162,12 @@ func resourceAlkiraInternetApplication() *schema.Resource {
 				},
 				Optional: true,
 			},
+			"ilb_credential_id": {
+				Description: "Credential ID of the ILB Target.  This can only be used if the associated connector is an AWS_VPC",
```
spikewang, line 166: where this credential comes from? It's AWS credential? or which credential. The comment is confusing.

alkira/resource_alkira_internet_applications.go
```diff
@@ -162,6 +162,12 @@ func resourceAlkiraInternetApplication() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "",
```
spikewang, line 169: I don't think default "" is acceptable.

vendor/github.com/alkiranet/alkira-client-go/alkira/internet_applications.go
spikewang: Also new optional field should be always `omitempty` to ensure backward compatibility.

alkira/resource_alkira_internet_applications.go
```diff
@@ -162,6 +162,11 @@ func resourceAlkiraInternetApplication() *schema.Resource {
 				},
 				Optional: true,
 			},
+			"ilb_credential_id": {
+				Description: "AWS Credential ID of the ILB Target.  This can only be used if the associated connector is an AWS_VPC",
```
spikewang, line 166: What exactly is an ILB? I don't think AWS has this concept at all, so not sure what is the credential for exactly.

spikewang (approved):

PR #354, opened 2025-02-13
AK-52645 - Add `implicit_group_id` to F5 service.

asim-ak:
Add `implicit_group_id` to F5 service.


spikewang (approved):
We could get this in now. it's read only field, so worst case scenario, it's just empty value :) @asim-ak

spikewang:
Please update the doc as well to pick up the latest.

spikewang (approved):

PR #355, opened 2025-03-01
AK-53221 - remove OverlappingRouteValidationEnabled in alkira_segment

asim-ak:
https://alkiranet.atlassian.net/browse/AK-53221


spikewang:
This is not used anywhere... we could wait till the next alkira-client-go pick to get it over.

spikewang:
I think we could close this PR and re-work it.

PR #357, opened 2025-03-04
AK-53294 Add optional field to enable route validation in resource segment

spikewang:
Add optional field enable_overlapping_route_validation in resource segment. This field should be available globally now and the default value will be `true`.


spikewang:
Tested on PROD and PREPROD and everything seems like fine.

PR #358, opened 2025-03-05
AK-50359 Fixing the initiator field description in IPsec and Advanced…

sankeerthmogili:
… IPsec connector


spikewang (approved):
I see thanks!

PR #359, opened 2025-03-10
AK-53131 - Add support for `fallback_to_tcp` in remote access connector

asim-ak:


alkira/resource_alkira_connector_remote_access.go
```diff
@@ -123,6 +123,12 @@ func resourceAlkiraConnectorRemoteAccess() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"fallback_to_tcp": {
+				Description: "Fallback to TCP when UDP fails." +
+					"Default value is `false`.",
+				Type:     schema.TypeBool,
+				Optional: true,
```
spikewang, line 130: Add a default value to make it clear. Even it is in the payload.

PR #364, opened 2025-04-02
AK-51518 - add instance level `ipsec_customer_gateway` (segment options) to Azure Express Route.

asim-ak:
Also update `segment_options` to take `segment_id` instead of `segment_name` (AK-54158)


spikewang:
@asim-ak don't forget the client-go changes. Push it in

alkira/resource_alkira_connector_azure_expressroute_helper.go
```diff
@@ -74,3 +184,80 @@ func expandAzureExpressRouteSegments(seg []interface{}, m interface{}) ([]alkira
...
+func flattenInstanceSegmentOptions(segmentOptions []alkira.InstanceSegmentOption, m interface{}) []interface{} {
+	if segmentOptions == nil {
+		return nil
+	}
+
+	result := make([]interface{}, len(segmentOptions))
+	for i, segOpt := range segmentOptions {
+		segmentId, err := getSegmentIdByName(segOpt.SegmentName, m)
+		if err != nil {
+			log.Printf("[DEBUG] Unable to find segment ID for name %s: %v", segOpt.SegmentName, err)
```
spikewang, line 212: This error message won't help much... No one will see it.. It's better to pop the `Error` up and then fail it, so return a error from the function and then fail the provider.

spikewang (changes requested):

asim-ak:
> @asim-ak don't forget the client-go changes. Push it in

https://github.com/alkiranet/alkira-client-go/pull/185

spikewang (dismissed):
Let's get in and try I guess. When initiator is `true`, usually, it may change... so not sure how it's handled in this case.

spikewang (approved):

PR #367, opened 2025-04-10
AK-54158 - fix data type for global `segment_id`

asim-ak:
add some logging.

### DIFF
--- a/alkira/resource_alkira_connector_azure_expressroute.go
+++ b/alkira/resource_alkira_connector_azure_expressroute.go
@@ -252,7 +252,7 @@ func resourceAlkiraConnectorAzureExpressRoute() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"segment_id": {
 							Description: "The ID of the segment.",
-							Type:        schema.TypeInt,
+							Type:        schema.TypeString,
 							Required:    true,
 						},
 						"customer_asn": {
--- a/alkira/resource_alkira_connector_azure_expressroute_helper.go
+++ b/alkira/resource_alkira_connector_azure_expressroute_helper.go
@@ -2,6 +2,7 @@ package alkira
 
 import (
 	"errors"
+	"log"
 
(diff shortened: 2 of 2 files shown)


alkira/resource_alkira_connector_azure_expressroute_helper.go
```diff
@@ -74,6 +75,7 @@ func expandInstanceSegmentOptions(in []interface{}, m interface{}) ([]alkira.Ins
 		if v, ok := segOptMap["segment_id"].(string); ok {
 			segmentName, err := getSegmentNameById(v, m)
 			if err != nil {
+				log.Printf("Unable to fetch segment_name for segment_id", v)
```
spikewang, line 78: This won't help much since no one will see it. The returned err should be on user's screen.

alkira/resource_alkira_connector_azure_expressroute.go
```diff
@@ -252,7 +252,7 @@ func resourceAlkiraConnectorAzureExpressRoute() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"segment_id": {
 							Description: "The ID of the segment.",
-							Type:        schema.TypeInt,
+							Type:        schema.TypeString,
```
spikewang, line 255: This is another mystery... I won't comment on this, but I saw either cases (actually one more weird case before).

spikewang (approved):

alkira/resource_alkira_connector_azure_expressroute.go
asim-ak: yeah , but since the segment helper functions take string I'll update the rest to also be of type string.

alkira/resource_alkira_connector_azure_expressroute_helper.go
asim-ak: yeah , it's only useful when running in TF_DEBUG mode.

PR #370, opened 2025-04-18
AK-54865 - remove default value for `ike_version`.

asim-ak:
If `ike_version` is sent in the payload, other fields like remote auth are required by the API.


spikewang (approved):
k, I am still not sure why this block has to be so random. To be honest, I would rather hide this field if it could only has one value, but I guess it won't help the problem you re trying to solve in this PR :)

PR #371, opened 2025-04-18
AK-53729 - add `scale_group_id` to AWS TGW connector.

asim-ak:


spikewang (approved):

spikewang (approved):
Don't forget to get the change in alkira-client-go

PR #372, opened 2025-04-22
AK-53137 Add initial support for probe resources 

asim-ak:
add support for TCP, HTTP, HTTPS probes in Alkira Terraform Provider.


alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+				Type:        schema.TypeList,
+				Required:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Description: "The type of network entity to probe." +
+								" Only `INTERNET_APPLICATION` supported for now.",
+							Type:     schema.TypeString,
+							Required: true,
```
spikewang, line 46: This is required field and it could be only 1 value. Let's just hide this one and assign it to `INTERNET_APPLICATION`.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+						},
+						"id": {
+							Description: "The ID of the network entity.",
```
spikewang, line 49: What ID is this? We never really have term `network entity` used anywhere else before... so it will be helpful to be more specific.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+					},
+				},
+			},
+			"uri": {
+				Description: "The URI to probe.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"headers": {
+				Description: "HTTP headers to include in the request.",
```
spikewang, line 62: This will be a specific format... We may mention this a bit here, or if it's too long, let's update the doc to explain this.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+							ValidateFunc: validation.StringInSlice([]string{"HTTP_STATUS_CODE", "HTTP_RESPONSE_BODY"}, false),
+						},
+						"status_code": {
+							Type:          schema.TypeString,
+							Description:   "Response code the response should have.",
+							Optional:      true,
+							ConflictsWith: []string{"regex"},
+						},
+						"regex": {
+							Description:   "Regex the response body should match.",
```
spikewang, line 87: We need to mention this in the description if it's conflict with status_code.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+							ConflictsWith: []string{"status_code"},
+							Optional:      true,
+						},
+					},
+				},
+			},
+			"failure_threshold": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The number of consecutive failures required to mark the probe as failed.",
```
spikewang, line 98: so we don't have a default value for this optional field? What if we don't set it... it will never fail? Then we should document it.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+			},
+			"success_threshold": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The number of consecutive successes required to mark the probe as successful.",
+			},
+			"period_seconds": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "How often (in seconds) to perform the probe.",
```
spikewang, line 108: All those I think we will have default value... We need to put in default value and explain the behaviour.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+			},
+			"timeout_seconds": {
```
spikewang, line 110: same as above

templates/resources/probe_http.md.tmpl
```diff
@@ -0,0 +1,31 @@
...
+
+{{ .Description | trimspace }}
+
+The `alkira_probe_http` resource allows you to configure and manage HTTP probes in the Alkira network.
+HTTP probes help monitor the health and availability of services by periodically sending HTTP requests 
+to specified endpoints and evaluating the responses based on configurable validation criteria.
+Connection will always be initiated to the EIPs associated with the Internet Application, 
+the host specified in the uri will be added to the host header
+
+~>**NOTE:**   This is a non-provisionable resource.
```
spikewang, line 18: Is this true? Even it's true, we don't need to say it explicitly, like all other non-provision able resources like billing tags.

spikewang: If this doesn't need to prov, how we handle the dependency with the resource it's trying to probe... IFA needs prov... so I created IFA + PROBE... what will happen...

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return diag.Diagnostics{{
+			Severity: diag.Warning,
+			Summary:  "Failed to Get HTTP Probe",
+			Detail:   err.Error(),
+		}}
+	}
+
+	if probe.Type != "HTTP" {
+		return diag.Errorf("Retrieved probe is not of HTTP type")
```
spikewang, line 152: This error will be just confusing... I think we should just say type mis-match...

alkira/resource_alkira_probe_https.go
```diff
@@ -0,0 +1,207 @@
...
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Description: "The type of network entity to probe." +
+								" Only `INTERNET_APPLICATION` supported for now.",
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"id": {
```
spikewang, line 48: Looks like we could totally avoid this block. So far it will be only `network_entity_id` and it could be only IFA ID.

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+						},
+					},
+				},
+			},
+			"uri": {
+				Description: "The URI to probe.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"headers": {
```
akshaival, line 61: headers are not supported for phase 1. It will be added in phase 2

alkira/resource_alkira_probe_http.go
akshaival: failure threshold is not optional

akshaival: So if its not set, the default is 3
This is the current behavior

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,190 @@
...
+				},
+			},
+			"failure_threshold": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The number of consecutive failures required to mark the probe as failed.",
+			},
+			"success_threshold": {
+				Type:        schema.TypeInt,
+				Optional:    true,
```
akshaival, line 102: default is 1

alkira/resource_alkira_probe_http.go
akshaival: default is 60

alkira/resource_alkira_probe_http.go
akshaival: default is 60

templates/resources/probe_http.md.tmpl
asim-ak: " IFA can be created and provisioned independently. For probe to be created it requires a provisioned IFA. It cannot be created for un-provisioned IFA. You can only create provisioned IFA+probe. Now once the probe is created for the provisioned IFA, IFA cannot be deleted until probe is deleted. We internally create this dependency here. For the IFA to be deleted now, probe will have to be deleted first."

alkira/resource_alkira_probe_http.go
asim-ak: Theres plan to add more types of network entities in next phases so we can't flatten this, we can set a default value for now.

alkira/resource_alkira_probe_http.go
asim-ak: apparently this is not supported yet so I've commented it out for now.

alkira/resource_alkira_probe_https.go
```diff
@@ -0,0 +1,215 @@
...
+				Optional:    true,
+			},
+			"disable_cert_validation": {
+				Description: "Whether to disable certificate validation.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+			},
+			"ca_certificate": {
+				Description: "Required when certificate validation" +
```
akshaival, line 73: Add string `ca certificate should have SAN extensions`

alkira/resource_alkira_probe_http.go
```diff
@@ -0,0 +1,198 @@
...
+				Description: "Validators for the HTTP response.",
+				Type:        schema.TypeList,
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Description: "Type of Validator, can be one of " +
+								"`HTTP_STATUS_CODE` or `HTTP_RESPONSE_BODY`.",
+							Type:         schema.TypeString,
+							Required:     true,
```
akshaival, line 78: UI needs atleast one of STATUS_CODE or RESPONSE_BODY
There is no such requirement on the backend side though. Are we following exactly how UI behaves? If yes we need to change this

alkira/resource_alkira_probe_https_helper.go
```diff
@@ -0,0 +1,154 @@
...
+	d.Set("server_name", params.ServerName)
+	d.Set("disable_cert_validation", params.DisableCertValidation)
+	d.Set("ca_certificate", params.CaCertificate)
+	// d.Set("headers", params.Headers)
+	d.Set("validators", flattenValidatorsHTTPS(params.Validators))
+
+	return nil
+}
+
+func expandNetworkEntityHTTPS(input []interface{}) *alkira.ProbeNetworkEntity {
```
akshaival, line 75: All these functions are same for all, cant we have a single one instead of 3 copies?

docs/resources/probe_http.md
```diff
@@ -0,0 +1,115 @@
...
+```terraform
+resource "alkira_probe_http" "basic_http" {
+  name = "basic-http-probe"
+
+  network_entity {
+    type = "INTERNET_APPLICATION"
+    id   = alkira_internet_application.example_application.id
+  }
+
+  uri = "www.alkira.net/api/health"
```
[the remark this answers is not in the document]
akshaival: http://www.alkira.net/api/health

this is also valid one

docs/resources/probe_http.md
```diff
@@ -0,0 +1,115 @@
...
+- `network_entity` (Block List, Min: 1, Max: 1) Network entity configuration. (see [below for nested schema](#nestedblock--network_entity))
+- `uri` (String) The URI to probe.
+
+### Optional
+
+- `enabled` (Boolean) Whether the probe is enabled.
+- `failure_threshold` (Number) The number of consecutive failures required to mark the probe as failed. Default is `3`, and the maximum value allowed is `50`.
+- `period_seconds` (Number) How often (in seconds) to perform the probe. Default value is `60`, and the maximum value allowed is `360`.
+- `success_threshold` (Number) The number of consecutive successes required to mark the probe as successful. Default value is `1`, and the maximum value allowed is `50`.
+- `timeout_seconds` (Number) Number of seconds after which the probe times out. Default value is `60`, and the maximum value allowed is `360`. `timeout_seconds` should always be greater than `period_seconds`.
```
akshaival, line 87: should always be less than or equal to period_seconds

examples/resources/alkira_probe_http/resource1.tf
```diff
@@ -0,0 +1,32 @@
...
+  enabled = true
+
+  network_entity {
+    type = "INTERNET_APPLICATION"
+    id   = alkira_internet_application.example_application.id
+  }
+
+  uri = "www.alkira.net/api/health"
+
+  headers = {
```
akshaival, line 12: need to comment this out for now

examples/resources/alkira_probe_https/resource1.tf
```diff
@@ -0,0 +1,24 @@
...
+
+  network_entity {
+    type = "INTERNET_APPLICATION"
+    id   = alkira_internet_application.example_application.id
+  }
+
+  uri         = "www.alkira.net/secure/endpoint"
+  server_name = "api.example.com"
+
+  ca_certificate = file("${path.module}/certs/exmaple_ca.pem")
```
akshaival, line 13: content of a file

docs/resources/probe_http.md
asim-ak: image

akshaival: Yes this needs to be updated

spikewang (approved):
Cool, let's try and see

PR #387, opened 2025-06-18
AK-57241 - support prefix and prefix range description in policy_prefix_list

asim-ak:


alkira/resource_alkira_policy_prefix_list.go
```diff
@@ -220,10 +237,11 @@ func generatePolicyPrefixListRequest(d *schema.ResourceData, m interface{}) (*al
 	}
 
 	list := &alkira.PolicyPrefixList{
-		Description:  d.Get("description").(string),
-		Name:         d.Get("name").(string),
-		Prefixes:     convertTypeSetToStringList(d.Get("prefixes").(*schema.Set)),
-		PrefixRanges: prefixRanges,
+		Description:   d.Get("description").(string),
+		Name:          d.Get("name").(string),
+		Prefixes:      extractPrefixesFromSchema(d),
```
spikewang, line 242: I think you could use a single function to do this and return `prefixes` and `prefixDetails`. It will be cleaner. Function name could be like this: `expandPrefixes` as our convention to expand a block in TF, that returns both struct...

so

```
prefixes, prefixDetails, err := expandPrefixes()

if err != nil...
```

alkira/resource_alkira_policy_prefix_list_helper.go
```diff
@@ -47,9 +48,49 @@ func expandPrefixListPrefixRanges(in []interface{}) ([]alkira.PolicyPrefixListRa
...
+				details[prefix] = &alkira.PolicyPrefixListDetails{Description: desc}
+			}
+		}
+	}
+	return details
+}
+
+func setPrefixesWithDescriptions(d *schema.ResourceData, prefixes []string, details map[string]*alkira.PolicyPrefixListDetails) {
+	var prefixList []map[string]interface{}
+	for _, p := range prefixes {
```
spikewang, line 88: I think we could do a quick test of performance to see this case... What's the max number of prefixes we could have? Any customers with big number of prefixes?

asim-ak: image

spikewang (approved):

spikewang:
what a hack, thanks @asim-ak

PR #575, opened 2026-03-23
[AK-65990] forces exportAllSubnets=false when vpc_subnet entries are …

jyoti-r27:
…present to prevent API 400 conflict

Fix GCP VPC connector 400 error when using vpc_subnet without explicit export_all_subnets

  The Alkira API rejects requests where exportAllSubnets=true and userInputPrefixes are both populated, as they are mutually exclusive. The provider was hardcoding ExportAllSubnets: true
  as a default and never clearing it when vpc_subnet blocks were present, causing all subnet-selection-mode connectors to fail with a 400 validation error unless the user explicitly wrote
   export_all_subnets = false in their HCL.

  Changes:
  - In expandGcpRouting(), after populating Prefixes from vpc_subnet blocks, explicitly set ExportAllSubnets = false when any prefixes are present — making the two fields mutually
  consistent before the request is sent.
  - In the GetRawConfig Create-time default block, added a len(Prefixes) == 0 guard so it no longer overrides the above correction back to true.
  - Updated TestExpandGcpRouting to replace the test case that was asserting the buggy behavior, and added three regression cases covering: gcp_routing block without export_all_subnets,
  no gcp_routing block at all, and multiple subnets — all with vpc_subnet present.

### DIFF
--- a/alkira/resource_alkira_connector_gcp_vpc_helper.go
+++ b/alkira/resource_alkira_connector_gcp_vpc_helper.go
@@ -45,6 +45,12 @@ func expandGcpRouting(in []interface{}, subnets *schema.Set) (*alkira.ConnectorG
 
 	exportOptions.Prefixes = prefixes
 
+	// Subnet entries and exportAllSubnets are mutually exclusive.
+	// If specific subnets are provided, exportAllSubnets must be false.
+	if len(prefixes) > 0 {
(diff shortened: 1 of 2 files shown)


camjrichards (approved):

asim-ak:
The changes look good, please do test the provider on the live environment.

amitmandke (approved):

PR #576, opened 2026-03-24
Add segment_ids validation for Bluecat service to prevent 500 error

asim-ak:
## Summary

- Add `MinItems: 1` validation to `segment_ids` schema field in Bluecat service
- Backend requires at least 1 segment (`@SoftLimit(min = 1, max = 1)`)
- Previously, empty `segment_ids` sent `null` for segments to the API, causing HTTP 500 after ~5 minutes of retries
- Now Terraform rejects empty `segment_ids` at plan time with a clear validation error

## Changes

- Add `MinItems: 1` to `segment_ids` schema definition
- Add unit tests verifying the MinItems constraint

## Testing

- Unit tests pass
- Lint passes
- Build succeeds

## Jira Ticket

AK-65889

### DIFF
--- a/alkira/resource_alkira_service_bluecat.go
+++ b/alkira/resource_alkira_service_bluecat.go
@@ -236,6 +236,7 @@ func resourceAlkiraBluecat() *schema.Resource {
 				Description: "IDs of segments associated with the service.",
 				Type:        schema.TypeSet,
 				Required:    true,
+				MinItems:    1,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
 			"service_group_name": {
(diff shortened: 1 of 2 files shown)


alkira/resource_alkira_service_bluecat.go
```diff
@@ -236,6 +236,7 @@ func resourceAlkiraBluecat() *schema.Resource {
 				Description: "IDs of segments associated with the service.",
 				Type:        schema.TypeSet,
 				Required:    true,
+				MinItems:    1,
```
sankeerthmogili, line 239: shouldn't we add MaxItems as well given backend constraints it to only 1?

snehachandrak: That is enforced as a soft limit by the API, although in practice we may never support more than one segment on the service.

snehachandrak (dismissed):

sankeerthmogili (dismissed):

sankeerthmogili (approved):

snehachandrak (approved):

PR #580, opened 2026-03-26
AK-66113: Fix export_all_subnets regression when vpc_subnet blocks removed

asim-ak:
## Summary

- Fix regression where removing all `vpc_subnet` blocks from a GCP VPC connector config caused a 400 API error
- Add bidirectional enforcement in `expandGcpRouting`: subnets absent now forces `exportAllSubnets=true` (the missing direction)
- Add `CustomizeDiff` validation for explicit `export_all_subnets=false` without `vpc_subnet` entries
- Update unit tests to reflect the corrected validation behavior

## Problem

When a GCP VPC connector was created with `vpc_subnet` blocks, `export_all_subnets` was saved as `false` in state. When the user later removed all `vpc_subnet` blocks from config (without explicitly setting `export_all_subnets=true`), the provider sent:

```
{exportAllSubnets: false, userInputPrefixes: []}
```

The API rejected this with:
```
400: userInputPrefixes cannot be empty when exportAllSubnets is FALSE
```

The existing code only handled one direction of the mutual exclusion:
- ✅ `vpc_subnet` present → force `exportAllSubnets=false`
- ❌ `vpc_subnet` absent → force `exportAllSubnets=true` (was missing)

## Solution

**`resource_alkira_connector_gcp_vpc_helper.go`** — Fix `expandGcpRouting` to enforce both directions unconditionally:

```go
if len(prefixes) > 0 {
    exportOptions.ExportAllSubnets = false  // subnets present → false
} else {
    exportOptions.ExportAllSubnets = true   // subnets absent  → true (the fix)
}
```

**`resource_alkira_connector_gcp_vpc.go`** — Extend `CustomizeDiff` to catch the invalid case of explicit `export_all_subnets=false` without `vpc_subnet` at plan time. Uses `GetRawConfig` to distinguish "user explicitly set false" from "computed from state".

## Changes

- `alkira/resource_alkira_connector_gcp_vpc.go` — Add `cty` import; extend `CustomizeDiff` to validate explicit `false` without subnets; use `GetRawConfig` to detect user-set vs computed value
- `alkira/resource_alkira_connector_gcp_vpc_helper.go` — Add `else` branch to force `ExportAllSubnets=true` when no subnets
- `alkira/resource_alkira_connector_gcp_vpc_helper_test.go` — Update test expectation for `export_all_subnets=false` without `vpc_subnet` to expect validation error

### Phase 1: Greenfield CRUD (live)

| Step | Result |
|------|--------|
| Create with `vpc_subnet` block | ✅ Connector created, `export_all_subnets=false` in state |
| Refresh | ✅ Plan shows **No changes** |
| **Remove `vpc_subnet` block (regression scenario)** | ✅ Update in-place (same ID), `export_all_subnets` → `true`, API payload `exportAllSubnets:true, userInputPrefixes:[]` |
| Idempotency after update | ✅ Plan shows **No changes** |
| Destroy | ✅ All resources cleaned up, state empty |

### Phase 8: Permutation Edge Cases (live)

| `export_all_subnets` | `vpc_subnet` | Valid? | Result |
|---------------------|--------------|--------|--------|
| `true` (explicit) | absent | ✅ Yes | Apply succeeds, state=`true` |
| `false` (explicit) | present | ✅ Yes | Apply succeeds, state=`false` |
| `true` (explicit) | present | ❌ No | `CustomizeDiff` rejects at plan: _"vpc_subnet cannot be specified when export_all_subnets is true"_ |
| omitted (computed) | present | ✅ Yes | Apply succeeds, state forced to `false` |
| omitted (computed) | absent | ✅ Yes | Apply succeeds, state defaults to `true` |
| `false` (explicit) | absent | ❌ No | `CustomizeDiff` rejects at plan: _"vpc_subnet must be specified when export_all_subnets is false"_ |

### Unit Tests

- `TestExpandGcpRouting` — All cases pass including bidirectional enforcement
- `TestGenerateGCPUserInputPrefixes` — All cases pass
- `TestSetGcpRoutingOptions` — All cases pass
- `TestSetGcpVpcSubnets` — All cases pass
- `TestGcpVpcValidateExportAllSubnetsWithVpcSubnet` — Updated to reflect `false`+no subnets now errors

### Code Quality

- `make lint` — 0 issues
- `make build` — clean
- Unit tests — All pass

### DIFF
--- a/alkira/resource_alkira_connector_gcp_vpc.go
+++ b/alkira/resource_alkira_connector_gcp_vpc.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/alkiranet/alkira-client-go/alkira"
+	"github.com/hashicorp/go-cty/cty"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -33,13 +34,30 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 				routing := gcpRouting.([]interface{})
 				if len(routing) > 0 {
 					routingCfg := routing[0].(map[string]interface{})
-					if exportAll, ok := routingCfg["export_all_subnets"].(bool); ok && exportAll {
-						if vpcSubnets, ok := d.GetOk("vpc_subnet"); ok {
-							if vpcSubnets.(*schema.Set).Len() > 0 {
-								return fmt.Errorf("vpc_subnet cannot be specified when export_all_subnets is true. " +
-									"When exporting all subnets, specific vpc_subnet entries should not be provided")
-							}
-						}
+					exportAll, _ := routingCfg["export_all_subnets"].(bool)
+
+					hasVpcSubnets := d.Get("vpc_subnet").(*schema.Set).Len() > 0
+
+					// Detect if export_all_subnets was explicitly written in config
+					// (as opposed to being computed from state)
+					exportAllExplicitlySet := false
+					rawConfig := d.GetRawConfig()
+					gcpRoutingRaw := rawConfig.GetAttr("gcp_routing")
+					if !gcpRoutingRaw.IsNull() && gcpRoutingRaw.IsKnown() && gcpRoutingRaw.LengthInt() > 0 {
+						exportAllRaw := gcpRoutingRaw.Index(cty.NumberIntVal(0)).GetAttr("export_all_subnets")
+						exportAllExplicitlySet = !exportAllRaw.IsNull()
+					}
+
+					// export_all_subnets=true WITH vpc_subnet entries is invalid
+					if exportAll && hasVpcSubnets {
+						return fmt.Errorf("vpc_subnet cannot be specified when export_all_subnets is true. " +
+							"When exporting all subnets, specific vpc_subnet entries should not be provided")
+					}
+
+					// export_all_subnets=false explicitly WITHOUT vpc_subnet entries is invalid
+					if exportAllExplicitlySet && !exportAll && !hasVpcSubnets {
+						return fmt.Errorf("vpc_subnet must be specified when export_all_subnets is false. " +
+							"Either set export_all_subnets to true or provide vpc_subnet entries")
 					}
 				}
 			}
--- a/alkira/resource_alkira_connector_gcp_vpc.go
+++ b/alkira/resource_alkira_connector_gcp_vpc.go
@@ -34,7 +34,7 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 				routing := gcpRouting.([]interface{})
 				if len(routing) > 0 {
 					routingCfg := routing[0].(map[string]interface{})
-					exportAll, _ := routingCfg["export_all_subnets"].(bool)
+					exportAll, ok := routingCfg["export_all_subnets"].(bool)
 
 					hasVpcSubnets := d.Get("vpc_subnet").(*schema.Set).Len() > 0
 
@@ -49,13 +49,13 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 					}
 
 					// export_all_subnets=true WITH vpc_subnet entries is invalid
-					if exportAll && hasVpcSubnets {
+					if ok && exportAll && hasVpcSubnets {
 						return fmt.Errorf("vpc_subnet cannot be specified when export_all_subnets is true. " +
 							"When exporting all subnets, specific vpc_subnet entries should not be provided")
 					}
 
 					// export_all_subnets=false explicitly WITHOUT vpc_subnet entries is invalid
-					if exportAllExplicitlySet && !exportAll && !hasVpcSubnets {
+					if ok && exportAllExplicitlySet && !exportAll && !hasVpcSubnets {
 						return fmt.Errorf("vpc_subnet must be specified when export_all_subnets is false. " +
 							"Either set export_all_subnets to true or provide vpc_subnet entries")
 					}
--- a/alkira/resource_alkira_connector_gcp_vpc.go
+++ b/alkira/resource_alkira_connector_gcp_vpc.go
@@ -36,7 +36,10 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
 					routingCfg := routing[0].(map[string]interface{})
 					exportAll, ok := routingCfg["export_all_subnets"].(bool)
 
-					hasVpcSubnets := d.Get("vpc_subnet").(*schema.Set).Len() > 0
+					var hasVpcSubnets bool
+					if v, ok := d.GetOk("vpc_subnet"); ok {
+						hasVpcSubnets = v.(*schema.Set).Len() > 0
+					}
 
 					// Detect if export_all_subnets was explicitly written in config
 					// (as opposed to being computed from state)
--- a/alkira/resource_alkira_connector_gcp_vpc_helper.go
+++ b/alkira/resource_alkira_connector_gcp_vpc_helper.go
@@ -5,7 +5,6 @@ import (
 	"log"
 
 	"github.com/alkiranet/alkira-client-go/alkira"
-	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
@@ -30,10 +29,6 @@ func expandGcpRouting(in []interface{}, subnets *schema.Set) (*alkira.ConnectorG
 			if v, ok := cfg["custom_prefix"].(string); ok {
 				importOptions.RouteImportMode = v
 			}
-
-			if v, ok := cfg["export_all_subnets"].(bool); ok {
-				exportOptions.ExportAllSubnets = v
-			}
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


jyoti-r27 (dismissed):

alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -33,13 +34,30 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
...
 					routingCfg := routing[0].(map[string]interface{})
-					if exportAll, ok := routingCfg["export_all_subnets"].(bool); ok && exportAll {
-						if vpcSubnets, ok := d.GetOk("vpc_subnet"); ok {
-							if vpcSubnets.(*schema.Set).Len() > 0 {
-								return fmt.Errorf("vpc_subnet cannot be specified when export_all_subnets is true. " +
-									"When exporting all subnets, specific vpc_subnet entries should not be provided")
-							}
-						}
+					exportAll, _ := routingCfg["export_all_subnets"].(bool)
+
```
sankeerthmogili, line 38: **[M2] Defensive regression — discard drops the `ok` signal**

```go
exportAll, _ := routingCfg["export_all_subnets"].(bool)
```

The original code used `ok` as a guard (`if exportAll, ok := ...; ok && exportAll`). Silently discarding it means that if the type assertion ever fails (e.g. nil interface value from an unexpected schema path), `exportAll` becomes `false` silently and Case 2 below could fire incorrectly. The schema guarantees `TypeBool` so this is unlikely in practice, but it is a defensive regression from the prior pattern. Suggest at least keeping it for clarity:

```go
exportAll, exportAllOk := routingCfg["export_all_subnets"].(bool)
```

Then the guard variable is available if needed later in this block too.

alkira/resource_alkira_connector_gcp_vpc_helper.go
```diff
@@ -46,9 +46,14 @@ func expandGcpRouting(in []interface{}, subnets *schema.Set) (*alkira.ConnectorG
 	exportOptions.Prefixes = prefixes
 
 	// Subnet entries and exportAllSubnets are mutually exclusive.
-	// If specific subnets are provided, exportAllSubnets must be false.
+	// Subnets present → exportAllSubnets must be false.
+	// No subnets → exportAllSubnets must be true.
+	// This handles both Create and Update (RCA: state carried false into Update
+	// after vpc_subnet blocks were removed, causing API 400).
```
sankeerthmogili: **[M1] "RCA:" is process language, not code documentation**

```go
// This handles both Create and Update (RCA: state carried false into Update
// after vpc_subnet blocks were removed, causing API 400).
```

"RCA" is incident/ticket terminology. Source comments should document the invariant, not the historical investigation. Suggest replacing with something like:

```go
// Enforced bidirectionally: if subnets are present exportAllSubnets must be
// false; if absent it must be true. Without the else branch, a stale state
// value of false is sent to the API when vpc_subnet blocks are removed,
// causing a 400 error.
```

alkira/resource_alkira_connector_gcp_vpc_helper_test.go
```diff
@@ -1045,26 +1046,31 @@ func validateExportAllSubnetsWithVpcSubnet(d *schema.ResourceData, client *alkir
...
-	exportAll, ok := routingCfg["export_all_subnets"].(bool)
-	if !ok || !exportAll {
-		return nil
-	}
+	exportAll, exportAllOk := routingCfg["export_all_subnets"].(bool)
 
-	// If export_all_subnets is true, vpc_subnet must be empty
-	vpcSubnets := d.Get("vpc_subnet")
-	if vpcSubnets == nil {
-		return nil
```
sankeerthmogili, line 1056: **[I1] `exportAllExplicitlySet` is always `true` when `gcp_routing` is present — the guard is non-functional in tests**

```go
exportAll, exportAllOk := routingCfg["export_all_subnets"].(bool)
exportAllExplicitlySet := exportAllOk
```

For a `TypeBool Optional+Computed` field, `schema.ResourceData.Get()` always returns a typed `false` — never a nil interface — so the type assertion succeeds and `exportAllOk` is `true` whenever `gcp_routing` is present, regardless of whether the user actually wrote `export_all_subnets` in their HCL.

The production `CustomizeDiff` correctly uses `GetRawConfig()` + `IsNull()` to distinguish "user explicitly wrote false" from "computed/state carried false". The test helper cannot replicate that with a plain `map[string]interface{}` — the two code paths are not equivalent.

Practical impact: in tests, Case 2 (`exportAllOk && !exportAll && !hasVpcSubnets`) will fire whenever `gcp_routing` is present and `vpc_subnet` is empty, even for configurations where production `CustomizeDiff` would correctly stay silent (because `export_all_subnets` was omitted and is null in raw config).

Consider adding an acceptance test that exercises the real `CustomizeDiff`, or at minimum add a comment to the test helper documenting that `exportAllOk` here does **not** match `!IsNull()` semantics and will always be true, so Case 2 coverage must come from acceptance tests.

sankeerthmogili (changes requested):
## Review summary

**Overall: Needs Work** — the core fix is correct, but there is dead code and a test accuracy issue to address before merging.

### Dead code not touched by this diff (must fix)

**C1** —  lines 34–35: the `if v, ok := cfg["export_all_subnets"].(bool)` block in `expandGcpRouting` is now unconditionally overridden by the new `if len(prefixes) > 0 / else` branches. Whatever value it reads from the config map is immediately stomped. Remove it.

**C2** —  lines 166–172: the `if d.Id() == ""` / `GetRawConfig` Create-only block in `generateConnectorGcpVpcRequest` that sets `ExportAllSubnets = true` when the field is null is now dead — `expandGcpRouting` already handles this path correctly via the `else` branch. Remove it or explicitly document why it must remain alongside the new logic.

### Missing test case (should fix)

No unit test covers the exact regression scenario: `gcp_routing` present with `export_all_subnets=false` (carried from state) + empty subnets after removal → `expandGcpRouting` must return `ExportAllSubnets=true`. Please add this test case to `TestExpandGcpRouting`.

alkira/resource_alkira_connector_gcp_vpc.go
```diff
@@ -33,13 +34,30 @@ func resourceAlkiraConnectorGcpVpc() *schema.Resource {
...
-							if vpcSubnets.(*schema.Set).Len() > 0 {
-								return fmt.Errorf("vpc_subnet cannot be specified when export_all_subnets is true. " +
-									"When exporting all subnets, specific vpc_subnet entries should not be provided")
-							}
-						}
+					exportAll, ok := routingCfg["export_all_subnets"].(bool)
+
+					hasVpcSubnets := d.Get("vpc_subnet").(*schema.Set).Len() > 0
+
+					// Detect if export_all_subnets was explicitly written in config
```
sankeerthmogili, line 44: **Potential nil pointer dereference.** If `d.Get("vpc_subnet")` returns `nil`, the type assertion `(*schema.Set)` will panic.

Safer alternative:
```go
var hasVpcSubnets bool
if v, ok := d.GetOk("vpc_subnet"); ok {
    hasVpcSubnets = v.(*schema.Set).Len() > 0
}
```

sankeerthmogili (commented):
Code review — the core fix is correct and well-targeted. A few issues noted inline.

alkira/resource_alkira_connector_gcp_vpc_helper.go
```diff
@@ -46,9 +41,16 @@ func expandGcpRouting(in []interface{}, subnets *schema.Set) (*alkira.ConnectorG
...
+	// Subnets present → exportAllSubnets must be false.
+	// No subnets → exportAllSubnets must be true.
+	// Enforced bidirectionally: if subnets are present exportAllSubnets must be
+	// false; if absent it must be true. Without the else branch, a stale state
+	// value of false is sent to the API when vpc_subnet blocks are removed,
+	// causing a 400 error.
 	if len(prefixes) > 0 {
 		exportOptions.ExportAllSubnets = false
+	} else {
+		exportOptions.ExportAllSubnets = true
```
sankeerthmogili, line 53: **Verify existing test compatibility.** Removing the explicit `export_all_subnets` read from the config map (the old `cfg["export_all_subnets"]` block) means `ExportAllSubnets` is now derived entirely from subnet presence. The existing test case "default route mode with subnets but export_all_subnets defaults to true" (~line 521 in the test file) expects `ExportAllSubnets: true` when subnets are present, which will now fail since this branch unconditionally sets `false`. That test expectation needs updating. Please confirm by running the full test suite.

alkira/resource_alkira_connector_gcp_vpc_helper_test.go
```diff
@@ -1045,26 +1071,39 @@ func validateExportAllSubnetsWithVpcSubnet(d *schema.ResourceData, client *alkir
...
-	}
+	// NOTE: The test helper cannot replicate CustomizeDiff's GetRawConfig() + IsNull()
+	// semantics for distinguishing "user explicitly wrote false" from "computed/state carried false".
+	// For a TypeBool Optional+Computed field, schema.ResourceData.Get() always returns a typed
+	// false (never nil), so exportAllOk is true whenever gcp_routing is present, regardless of
+	// whether the user actually wrote export_all_subnets in their HCL. The production CustomizeDiff
+	// correctly uses GetRawConfig() to detect explicit writes. This limitation means Case 2 below
+	// may fire in unit tests for configs where production CustomizeDiff would correctly stay silent.
```
sankeerthmogili, line 1082: Nit: `exportAllExplicitlySet := exportAllOk` is broader than what the production `CustomizeDiff` does with `GetRawConfig`. The comment above explains this well — consider adding a `// TODO:` referencing acceptance test coverage for Case 2 so it doesn't get lost.

asim-ak: Acceptance tests cannot test plan-time validation errors (Case 2 is supposed to  fails at terraform plan, blocking terraform apply). Manual testing validates the CustomizeDiff behavior. The test helper already documents the limitation that it cannot replicate GetRawConfig() + IsNull() semantics.

alkira/resource_alkira_connector_gcp_vpc_helper.go
asim-ak: All tests pass locally as well as on the PR

sankeerthmogili (approved):

PR #604, opened 2026-04-08
AK-66046: Fix backward compatibility for deprecated prefixes field in policy_prefix_list

asim-ak:
## Summary

- Use `GetRawConfig` in Read and generateRequest to detect which field the user's HCL declares
- Add `GetOk` fallback for import/refresh where raw config is unavailable
- Remove `Computed` from `prefixes` to eliminate post-import drift
- Clear stale cross-field state on both Read paths
- Mark `prefixes` as deprecated with `ConflictsWith` against `prefix`

## Problem

Customers upgrading from older provider versions (v1.3.x) that used the `prefixes` field hit two breaking issues:

1. **Read always populated `prefix` blocks** even when user config only had `prefixes`
2. **`generatePolicyPrefixListRequest` hard-rejected `prefixes`** — blocked all creates/updates

Additionally, using `GetOk` to detect which field to populate checked state, not config. Customers could never migrate from `prefixes` to `prefix` blocks because stale state kept Read on the wrong path.

### Read: GetRawConfig with fallback
- Uses `GetRawConfig` to check what the user's HCL actually declares
- Falls back to `GetOk` for import/refresh where raw config isn't available
- Deprecated path: populates `prefixes`, clears stale `prefix` blocks
- Prefix block path: populates `prefix` blocks, does not touch `prefixes`

### generateRequest: GetRawConfig with fallback
- Same pattern with `IsKnown`/`IsNull` guards to prevent panics
- Falls back to `GetOk` when raw config is unavailable (C3 fix)

### Schema: Removed Computed from prefixes
- Eliminates post-import drift (`prefixes = (known after apply)`)
- `prefixes` is now purely `Optional` + `Deprecated`

### Stale state cleanup
- Deprecated path clears `prefix` blocks from state
- Empty API response clears `prefixes` (reflects API truth)

## Changes

- `alkira/resource_alkira_policy_prefix_list.go` — GetRawConfig in Read, stale state cleanup, remove Computed
- `alkira/resource_alkira_policy_prefix_list_helper.go` — GetRawConfig with fallback in generateRequest
- `alkira/resource_alkira_policy_prefix_list_helper_test.go` — 8 new unit tests

## Unit Tests

- `TestReadFieldSelection_FallbackToState` (5 sub-tests): state-based field selection, empty API clearing, stale cross-field cleanup
- `TestGenerateRequest_PrefixBlockPath` (3 sub-tests): prefix block API request, empty request, combined with prefix_range
- All 31 prefix list unit tests pass

[...]

PR #610, opened 2026-04-11
AK-66870: Fix AWS VPC connector vpc_cidr to vpc_subnet mode switch

asim-ak:
## Summary

- Remove `Computed: true` from `vpc_cidr` and `vpc_subnet` schema fields to fix silent mode switch failure
- Retain `Computed: true` on `vpc_route_table` and `overlay_subnets` (no `ConflictsWith`, backend auto-populates)

## Problem

When updating an AWS VPC connector from `vpc_cidr` mode to `vpc_subnet` mode, the API receives stale CIDR prefixes instead of the new subnet prefixes. The connector silently stays in CIDR mode — no error, no drift warning.

**Root cause:** `expandUserInputPrefixes()` checks `len(cidr) > 0` first and returns CIDR data immediately, never reaching the subnet branch. With `Computed: true` on both fields, `d.Get("vpc_cidr")` returns stale state values even after the user removes `vpc_cidr` from config. The stale CIDRs short-circuit the expand function.

**Introduced by:** AK-65432 (Read populates routing fields) + AK-66346 (added `Computed: true` to prevent drift). Neither is in the published registry version — only on `dev`.

## Solution

Remove `Computed: true` from `vpc_cidr` and `vpc_subnet`. Without `Computed`, when the user switches modes:
- Terraform clears the removed field to its zero value in the plan
- `d.Get("vpc_cidr")` returns `[]` (empty)
- `expandUserInputPrefixes` reaches the subnet branch
- Correct `SUBNET` payload sent to API

This is safe because `expandUserInputPrefixes` requires at least one of the two fields, so users must always specify one in config. This matches the `connector_oci_vcn` pattern and the approach in PR #604 for `policy_prefix_list`.

## Changes

- `alkira/resource_alkira_connector_aws_vpc.go` — Remove `Computed: true` from `vpc_cidr` and `vpc_subnet`

### Live Testing

**Test environment:** `test/AK-66870-vpc-cidr-subnet-mode-switch/`

- Greenfield vpc_cidr (full config) — API payload `type:CIDR`, plan "No changes"
- Greenfield vpc_subnet (full config, 2 subnets) — API payload `type:SUBNET`, plan "No changes"
- Greenfield minimal vpc_cidr (no route table, no overlay) — Backend defaults handled, no drift
- Greenfield minimal vpc_subnet (no route table, no overlay) — No drift
- **KEY TEST:** Mode switch CIDR → SUBNET — API payload `type:SUBNET` (not stale `type:CIDR`)
- **KEY TEST:** Mode switch SUBNET → CIDR — API payload `type:CIDR` (not stale `type:SUBNET`)
- Mode switch CIDR → 2 subnets — API payload 2x `type:SUBNET`, no drift
- Brownfield upgrade (registry → fixed build) — No connector changes on upgrade plan
- Import vpc_cidr with `generate-config-out` — Generated config correct, no drift

### Code Quality

- `make lint` — 0 issues
- `go build` — clean
- Unit tests — All pass

### DIFF
--- a/alkira/resource_alkira_connector_aws_vpc_helper.go
+++ b/alkira/resource_alkira_connector_aws_vpc_helper.go
@@ -198,23 +198,23 @@ func expandAwsVpcRouteTables(in *schema.Set) []alkira.RouteTables {
 // expandUserInputPrefixes generate UserInputPrefixes used in AWS-VPC connector
 func expandUserInputPrefixes(cidr []interface{}, subnets *schema.Set, overlaySubnets []interface{}) ([]alkira.InputPrefixes, error) {
 
-	if len(cidr) == 0 && subnets == nil {
-		return nil, fmt.Errorf("ERROR: either \"vpc_subnet\" or \"vpc_cidr\" must be specified")
-	}
+	hasCidr := len(cidr) > 0
+	hasSubnets := subnets != nil && subnets.Len() > 0
 
 	// Processing overlay_subnets
 	log.Printf("[DEBUG] Processing overlay_subnets %v", overlaySubnets)
 	overlaySubnetList := make([]alkira.InputPrefixes, len(overlaySubnets))
 
-	if len(overlaySubnets) > 0 {
-		for i, value := range overlaySubnets {
-			overlaySubnetList[i].Value = value.(string)
-			overlaySubnetList[i].Type = "OVERLAY_SUBNETS"
-		}
+	for i, value := range overlaySubnets {
+		overlaySubnetList[i].Value = value.(string)
+		overlaySubnetList[i].Type = "OVERLAY_SUBNETS"
 	}
 
-	// Processing vpc_cidr
-	if len(cidr) > 0 {
+	switch {
+	case hasCidr && hasSubnets:
+		return nil, fmt.Errorf("ERROR: \"vpc_cidr\" and \"vpc_subnet\" cannot both be specified")
+
+	case hasCidr:
 		log.Printf("[DEBUG] Processing vpc_cidr %v", cidr)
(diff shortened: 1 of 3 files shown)


snehachandrak:
@asim-ak the actual root cause is in
`expandUserInputPrefixes()` checks `len(cidr) > 0` first and returns CIDR data immediately, never reaching the subnet branch. <---- This needs to be fixed.

snehachandrak (changes requested):
Fix the read function and expandUserInputPrefixes.

alkira/resource_alkira_connector_aws_vpc_helper.go
```diff
@@ -198,23 +198,23 @@ func expandAwsVpcRouteTables(in *schema.Set) []alkira.RouteTables {
...
+	for i, value := range overlaySubnets {
+		overlaySubnetList[i].Value = value.(string)
+		overlaySubnetList[i].Type = "OVERLAY_SUBNETS"
 	}
 
-	// Processing vpc_cidr
-	if len(cidr) > 0 {
+	switch {
+	case hasCidr && hasSubnets:
+		return nil, fmt.Errorf("ERROR: \"vpc_cidr\" and \"vpc_subnet\" cannot both be specified")
```
snehachandrak, line 215: why do we have this validation? the connector can have both cidrs and subnets.

asim-ak: We have an existing `ConflictsWith` on the two fields, this is just a defensive check for imports, to match the TF behavior(which drifts a bit form the UI/API, this will be taken up in the follow up fixes for this resource).

snehachandrak (approved):

PR #613, opened 2026-04-15
AK-67039: making akamai_bgp_authentication_key sensitive true and Wri…

yashgupta912121:
Tested the following
  Scenario 1 — Create flow still works                                                                                          
  - Applied alkira_connector_akamai_prolexic resource with akamai_bgp_authentication_key = "dummy-auth-key"                     
  - Result: Connector created successfully (id 17159). akamai_bgp_authentication_key shown as (write-only attribute) in plan,   
  confirming WriteOnly: true is honored. tunnel_configuration output populated correctly.                                       
                                                                                                                                
  Scenario 2 — Import + -generate-config-out (the actual bug repro)                                                          
  - Ran terraform import + terraform plan -generate-config-out=generated.tf against connector id 17159                          
  - Before fix: generated config had akamai_bgp_authentication_key = null (no sensitive marker) and plan failed                 
  - After fix: generated config correctly emits akamai_bgp_authentication_key = null # sensitive — matches the expected output  
  in the bug description                                                          
  
  Did this additional testing:
   - Scenario 1 — Create: Fresh terraform apply in sun-staging-prolexic/ → connector 17163 created, credential_id
  captured as baseline
  - Scenario 2 — Non-credential update: Changed size SMALL → MEDIUM and applied → credential_id unchanged; log: [INFO]
   Updating credential-akamai-prolexic                                                                                
  - Scenario 3 — Auth key update: Changed akamai_bgp_authentication_key from "dummy-auth-key" → "dummy-auth-key-v2"   
  and applied → credential_id unchanged; log: [INFO] Updating credential-akamai-prolexic                              
  - Scenario 4 — Import + generate-config regression: In sun-staging-prolexic-import/, ran terraform plan          
  -generate-config-out=generated.tf against id 17163 → generated.tf correctly emits akamai_bgp_authentication_key =   
  null # sensitive                                                           
  - Scenario 5 — Destroy + credential cleanup: terraform destroy → connector gone; log: [INFO] Deleting               
  credential-akamai-prolexic, no [WARN] Failed to delete credential

### DIFF
--- a/alkira/resource_alkira_connector_akamai_prolexic.go
+++ b/alkira/resource_alkira_connector_akamai_prolexic.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 
 	"github.com/alkiranet/alkira-client-go/alkira"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -55,9 +57,16 @@ func resourceAlkiraConnectorAkamaiProlexic() *schema.Resource {
 				Required:    true,
 			},
 			"akamai_bgp_authentication_key": {
-				Description: "The Akamai BGP Authentication Key.",
-				Type:        schema.TypeString,
-				Required:    true,
+				Description: "The Akamai BGP Authentication Key. " +
+					"It could also be set by environment variable " +
+					"`AK_AKAMAI_BGP_AUTHENTICATION_KEY`. This value is " +
+					"stored as a credential in the Alkira backend and " +
+					"is never returned by the API, so it is not " +
+					"persisted in Terraform state.",
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+				WriteOnly: true,
 			},
 			"byoip_options": {
 				Description: "BYOIP options.",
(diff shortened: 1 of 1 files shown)


AK-Benjamin:
Did you test the update flow?  It looks like we may be creating a new credential on every operation. 
Just make sure that on update the credentialId doesn't change (also check that a new credential is not created for the tenant)

alkira/resource_alkira_connector_akamai_prolexic.go
```diff
@@ -261,6 +271,10 @@ func resourceConnectorAkamaiProlexicRead(ctx context.Context, d *schema.Resource
 	d.Set("size", connector.Size)
 	d.Set("description", connector.Description)
 
+	// akamai_bgp_authentication_key is WriteOnly and never returned
+	// by the API; it is maintained in the user's HCL configuration
+	// and not persisted to state.
```
AK-Benjamin: comment no longer needed here

alkira/resource_alkira_connector_akamai_prolexic.go
```diff
@@ -401,21 +427,46 @@ func generateConnectorAkamaiProlexicRequest(d *schema.ResourceData, m interface{
...
+		BgpAuthenticationKey: bgpAuthKey,
 	}
 
 	client := m.(*alkira.AlkiraClient)
-	credentialId, err := client.CreateCredential(d.Get("name").(string), alkira.CredentialTypeAkamaiProlexic, c, 0)
+	credentialId := d.Get("credential_id").(string)
 
-	if err != nil {
-		return nil, err
+	if credentialId == "" {
```
AK-Benjamin, line 452: for connectors that were created before this change is released, this will create an extra credential entry.  Don't think it will cause any functional issues though but want to note in case it needs to be addressed.

AK-Benjamin (dismissed):

AK-Benjamin (approved):

PR #616, opened 2026-04-15
AK-66323 support inter cxp route policy resource

snehachandrak:


sankeerthmogili (commented):
## Code Review — AK-66323 Inter-CXP Route Policy Terraform Resource

1281 additions, 2 deletions across 20 files. Adds a new `alkira_policy_inter_cxp_routing` resource with CRUD operations, helper functions, comprehensive documentation with 11 example `.tf` files, and the vendored alkira-client-go API client.

Cross-referenced against the TPS backend model at `tenant-provisioning-service/src/main/java/com/alkira/tenantprovisioningservice/model/intercxproutepolicy/`.

---

### Critical Issues

**1. Action values in schema don't match examples — "DROP" vs "DENY"**

Schema validation (`resource_alkira_policy_inter_cxp_routing.go:127-128`):
```go
ValidateFunc: validation.StringInSlice([]string{"ALLOW", "DENY"}, false),
```

But multiple example files use `"DROP"` as the action:
- `resource_as_path_filter.tf:10` → `action = "DROP"`
- `resource_drop_all.tf:14` → `action = "DROP"`
- `resource_group_match.tf:22` → `action = "DROP"`
- `resource_multi_rule_pipeline.tf:11,28` → `action = "DROP"`
- `resource_prefix_filter.tf:17` → `action = "DROP"`

The TPS backend validates against `ALLOW` and `DENY` (from `InterCxpRoutePolicyRule.action`). **All "DROP" examples will fail validation.** Either the schema should accept "DROP" instead of "DENY", or all examples need to be updated to use "DENY".

---

**2. Missing `source_cxps` length validation — backend requires exactly 1**

The TPS backend enforces that `sourceCxps` must contain **exactly 1 element** (validated in `InterCxpRoutePolicyValidator`). The Terraform schema has no `MaxItems` constraint:

```go
"source_cxps": {
    Type:     schema.TypeList,
    Required: true,
    Elem:     &schema.Schema{Type: schema.TypeString},
},
```

Users can pass 0 or multiple source CXPs and will get a server-side validation error with no client-side feedback. Add `MaxItems: 1, MinItems: 1` to catch this early.

---

### Important Issues

**3. Inconsistent nil-zeroing across match list fields**

In `expandPolicyInterCxpRoutingRuleMatch` (helper file), some fields set `nil` when empty and others don't:

```go
// These nil-zero when empty:
match.PrefixListIds = convertTypeSetToIntList(v)
if len(match.PrefixListIds) == 0 { match.PrefixListIds = nil }  // ✓

// These DON'T nil-zero:
match.CommunityListIds = convertTypeSetToIntList(v)              // ✗
match.AsPathListIds = convertTypeSetToIntList(v)                 // ✗
```

If `convertTypeSetToIntList` returns an empty slice for unset fields, the JSON will serialize as `"communityListIds": []` instead of being omitted. The backend may reject or behave differently with an empty array vs. null/omitted.

**Recommendation**: Either nil-zero all list fields consistently, or none of them.

---

**4. No client-side validation for mutually exclusive match conditions**

The TPS backend enforces:
- If `match.all = true`, no other match conditions are allowed
- `connectorGroupIds` and `segmentResourceIds` are mutually exclusive
- If `action = DENY`, `set` params must be null

None of these are validated in the Terraform schema or `CustomizeDiff`. Users will only discover these constraints via server-side errors, which are harder to debug.

**Recommendation**: Add `ConflictsWith` or `CustomizeDiff` validation for at least the `match_all` exclusivity and `match_group_ids`/`match_segment_resource_ids` mutual exclusion.

---

**5. `segment_id` stored as string but backend uses segment name**

The resource stores `segment_id` (a string-typed ID) and converts to segment name via `getSegmentNameById` on write and `getSegmentIdByName` on read. However, `segment_id` is `schema.TypeString` — if the user provides an integer ID (e.g., from `alkira_segment.seg1.id` which returns a string representation of the numeric ID), this works. But if the segment name changes, the stored ID still maps correctly.

This is fine functionally, but the description says "ID of the segment" while the field accepts the string form. Consider adding `ForceNew: true` if the backend doesn't allow changing the segment after creation, or add a note that segment changes require policy recreation.

---

### Minor Issues

**6. Description typo**

`source_cxps` description (line 93): "Exactly one CXP is allowed The CXP must carry the policy segment." — missing period after "allowed".

---

**7. `match_all` schema description says "Exclusive" but no enforcement**

The description says "Exclusive — cannot be combined with any other match condition" but neither `ConflictsWith` nor any custom validation enforces this. The server will reject it, but the error message won't be as clear.

---

**8. `set_as_path_prepend` description inconsistency**

Description says "Space-separated values 0–65535. Example: `100 100 100`" but the TPS backend validates against a regex pattern that may be more restrictive. The description should match the actual validation.

---

### Strengths

- Excellent documentation with 11 real-world example `.tf` files covering minimal, filtering, prepending, community tagging, composite matching, disabled, drop-all, extended community, group match, multi-rule pipeline, and prefix filter scenarios
- Clean client-go struct design matching the TPS backend model exactly
- Proper use of `AlkiraAPI[T]` generics for the client (no boilerplate CRUD)
- `CustomizeDiff` for provision state recovery (FAILED → SUCCESS) follows established pattern
- Read function correctly converts between segment name (API) and segment ID (Terraform state)
- Proper `omitempty` on JSON tags and nil-check for `Set` field in the rule model

---

### Assessment

**Not ready to merge** — The "DROP" vs "DENY" mismatch (issue #1) means all examples will fail at runtime. The missing `source_cxps` cardinality validation (issue #2) will cause confusing server errors. The inconsistent nil-zeroing (issue #3) could cause unexpected API behavior.

alkira/resource_alkira_policy_inter_cxp_routing.go
```diff
@@ -0,0 +1,384 @@
...
+							MaxItems:    1,
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+							Optional:    true,
+						},
+						"match_community_list_ids": {
+							Description: "IDs of Community Lists to match.",
+							Type:        schema.TypeSet,
+							MaxItems:    1,
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+							Optional:    true,
```
sankeerthmogili, line 142: **Critical**: Schema validates `["ALLOW", "DENY"]` but all example `.tf` files use `"DROP"` instead of `"DENY"`. The TPS backend enum is `ALLOW`/`DENY`. Either:
1. Change all examples from `"DROP"` to `"DENY"`, or
2. If the API actually accepts `"DROP"`, add it to the validation list

Files affected: `resource_as_path_filter.tf`, `resource_drop_all.tf`, `resource_group_match.tf`, `resource_multi_rule_pipeline.tf`, `resource_prefix_filter.tf`

alkira/resource_alkira_policy_inter_cxp_routing.go
```diff
@@ -0,0 +1,384 @@
...
+				Type:     schema.TypeList,
+				MinItems: 1,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Description: "The name of the rule. Must be unique within the policy.",
+							Type:        schema.TypeString,
+							Required:    true,
+						},
```
sankeerthmogili, line 110: **Critical**: Missing `MaxItems: 1, MinItems: 1`. The TPS backend enforces that `sourceCxps` must contain **exactly 1 element**. Without this constraint, users can pass multiple CXPs and get a confusing server-side validation error.

```go
"source_cxps": {
    Type:     schema.TypeList,
    Required: true,
    MinItems: 1,
    MaxItems: 1,
    Elem:     &schema.Schema{Type: schema.TypeString},
},
```

alkira/resource_alkira_policy_inter_cxp_routing_helper.go
```diff
@@ -0,0 +1,156 @@
...
+	if v, ok := in["match_community_list_ids"].(*schema.Set); ok {
+		match.CommunityListIds = convertTypeSetToIntList(v)
+	}
+	if v, ok := in["match_extended_community_list_ids"].(*schema.Set); ok {
+		match.ExtendedCommunityListIds = convertTypeSetToIntList(v)
+		if len(match.ExtendedCommunityListIds) == 0 {
+			match.ExtendedCommunityListIds = nil
+		}
+	}
+	if v, ok := in["match_as_path_list_ids"].(*schema.Set); ok {
```
sankeerthmogili, line 86: **Important**: Inconsistent nil-zeroing. `PrefixListIds`, `ExtendedCommunityListIds`, `SegmentResourceIds`, and `ConnectorGroupIds` get nil-zeroed when empty, but `CommunityListIds` and `AsPathListIds` do **not**.

If `convertTypeSetToIntList` returns `[]int{}` for unset fields, the JSON will send `"communityListIds": []` instead of omitting the field. The backend may treat these differently.

Apply the same pattern to all list fields for consistency.

alkira/resource_alkira_policy_inter_cxp_routing.go
```diff
@@ -0,0 +1,384 @@
...
+				Required: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"rule": {
+				Type:     schema.TypeList,
+				MinItems: 1,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
```
sankeerthmogili, line 106: **Minor**: Typo — "Exactly one CXP is allowed The CXP must carry" → missing period: "Exactly one CXP is allowed**.** The CXP must carry"

alkira/resource_alkira_policy_inter_cxp_routing.go
```diff
@@ -0,0 +1,384 @@
...
+						"match_community_list_ids": {
+							Description: "IDs of Community Lists to match.",
+							Type:        schema.TypeSet,
+							MaxItems:    1,
+							Elem:        &schema.Schema{Type: schema.TypeInt},
+							Optional:    true,
+						},
+						"match_extended_community_list_ids": {
+							Description: "IDs of Extended Community Lists to match.",
+							Type:        schema.TypeSet,
```
sankeerthmogili, line 147: **Important**: The description says "Exclusive — cannot be combined with any other match condition" but there's no `ConflictsWith` or `CustomizeDiff` validation enforcing this. The server will reject it, but the error will be a generic API error rather than a clear Terraform plan-time message.

Consider adding validation in `CustomizeDiff` that checks if `match_all = true` and any `match_*_ids` fields are also set.

sankeerthmogili:
## Review Findings (with TPS + client-go cross-check)

### Important: `ForceNew: true` missing on `direction` field

The `direction` field schema says "Immutable after creation" but does not set `ForceNew: true`. I checked the TPS `InterCxpRoutePolicyTranslator.updateEntity()` — it deliberately **does not** update `direction` on the existing entity. This means if a user changes `direction` in their Terraform config, the update API call silently ignores it (the direction stays as originally set). Terraform reports success, but the actual state doesn't match the config, causing **perpetual drift** on every subsequent `terraform plan`.

**Fix:** Add `ForceNew: true` to the `direction` schema attribute.

```go
"direction": {
    Description: "The direction of the policy. Only `OUTBOUND` is supported in Phase 1. " +
        "Immutable after creation.",
    Type:     schema.TypeString,
    Required: true,
    ForceNew: true,  // <-- add this
    ValidateFunc: validation.StringInSlice(
        []string{"OUTBOUND"}, false),
},
```

### Request: Add a test for `CustomizeDiff` validation

The `CustomizeDiff` function validates `match_all` exclusivity and other constraints using type assertions on `*schema.Set`. These assertions can silently fail depending on how Terraform SDK represents field values during the diff phase. Please add a test that exercises the `CustomizeDiff` validation — specifically the case where `match_all = true` is combined with a match field (e.g., `match_prefix_list_ids`). This confirms the `*schema.Set` assertion works at plan time and users get the expected error message instead of silently passing invalid configs through to the API.

### Minor: stale `go.sum` entries

Two intermediate `alkira-client-go` pseudo-versions are in `go.sum` but not referenced in `go.mod`. Running `go mod tidy` would clean these up.

### Confirmed non-issue: `match_all` bool behavior

Checked the TPS model — `InterCxpRoutePolicyRuleMatch.all` is `Boolean` with `@Builder.Default = false` and `@JsonInclude(NON_NULL)`. The validator treats `null` and `false` identically (`ObjectUtils.isEmpty(match.getAll()) || Boolean.FALSE.equals(match.getAll())`). The client-go `bool` type (always sends `false` when unset) is compatible. No issue here.

alkira/resource_alkira_policy_inter_cxp_routing.go
```diff
@@ -0,0 +1,433 @@
...
+		ReadContext:   resourcePolicyInterCxpRoutingRead,
+		UpdateContext: resourcePolicyInterCxpRoutingUpdate,
+		DeleteContext: resourcePolicyInterCxpRoutingDelete,
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, m interface{}) error {
+			client := m.(*alkira.AlkiraClient)
+
+			old, _ := d.GetChange("provision_state")
+
+			if client.Provision && old == "FAILED" {
+				d.SetNew("provision_state", "SUCCESS")
```
sankeerthmogili, line 29: **`ForceNew: true` needed here.**

The description says "Immutable after creation", but without `ForceNew` Terraform will attempt an in-place update. The TPS `InterCxpRoutePolicyTranslator.updateEntity()` deliberately skips updating `direction` — so the API silently ignores the change. Terraform reports success, but the actual state does not match the config, causing **perpetual drift** on every subsequent `terraform plan`.

```suggestion
				ForceNew: true,
				ValidateFunc: validation.StringInSlice(
```

alkira/resource_alkira_policy_inter_cxp_routing.go
```diff
@@ -0,0 +1,433 @@
...
+				if hasGroups && hasSegRes && groupIds.Len() > 0 && segResIds.Len() > 0 {
+					return fmt.Errorf("rule[%d] %q: match_group_ids and match_segment_resource_ids are mutually exclusive", i, rule["name"])
+				}
+
+				// match_group_ids and match_extended_community_list_ids are mutually exclusive.
+				extCommIds, hasExtComm := rule["match_extended_community_list_ids"].(*schema.Set)
+				if hasGroups && hasExtComm && groupIds.Len() > 0 && extCommIds.Len() > 0 {
+					return fmt.Errorf("rule[%d] %q: match_group_ids and match_extended_community_list_ids are mutually exclusive", i, rule["name"])
+				}
+
```
sankeerthmogili: **Add a test that exercises this `CustomizeDiff` validation.**

The `*schema.Set` type assertion here can silently return `ok == false` depending on how the SDK represents field values during the diff phase — if that happens, the validation is skipped and invalid configs pass through to the API (where the user gets a worse error message).

Please add a test case with `match_all = true` combined with a match field (e.g. `match_prefix_list_ids`) to confirm this assertion works at plan time.

alkira/resource_alkira_policy_inter_cxp_routing.go
snehachandrak: added validation instead of ForceNew, TPS API will also throw error validation.

sankeerthmogili (dismissed):

sankeerthmogili (approved):

PR #622, opened 2026-04-16
AK-66663: Add WriteOnly+Sensitive to alkira_service_infoblox credential fields

asim-ak:
## Summary

- Add `WriteOnly: true` + `Sensitive: true` to `shared_secret`, `grid_master.username`, `grid_master.password`, and `instance.password`
- Add `getInfobloxWriteOnlyValue()` and `getInfobloxNestedWriteOnlyValue()` helpers using `GetRawConfigAt()`
- Update `expandInfobloxGridMaster` and `expandInfobloxInstances` to read credentials from raw config via indexed cty paths

## Problem

Infoblox service credential fields (`shared_secret`, grid master credentials, instance passwords) were stored in plaintext in Terraform state. On import, Required WriteOnly fields caused errors because the API does not return credential values.

## Solution

Mark all 4 credential fields as `WriteOnly: true` + `Sensitive: true`. For the top-level `shared_secret`, use `getInfobloxWriteOnlyValue()` with `GetRawConfigAt`. For nested fields in `grid_master` and `instance` TypeList blocks, `d.Get()` returns empty strings for WriteOnly fields — fixed by passing `ResourceData` to expand functions and using `GetRawConfigAt` with indexed cty paths.

## Changes

- `alkira/resource_alkira_service_infoblox.go` — Add WriteOnly+Sensitive to 4 schema fields; update `generateInfobloxRequest` to use `getInfobloxWriteOnlyValue` for `shared_secret` and pass `d` to expand functions
- `alkira/resource_alkira_service_infoblox_helper.go` — Add `getInfobloxWriteOnlyValue()` and `getInfobloxNestedWriteOnlyValue()` helpers; update `expandInfobloxGridMaster` and `expandInfobloxInstances` to accept `ResourceData` and read WriteOnly fields from raw config
- `docs/resources/service_infoblox.md` — Regenerated docs

### Unit Tests

- `go test ./alkira/...` — All pass (0 failures)
- `make lint` — 0 issues
- `make build` — Clean

### Greenfield CRUD

**Test environment:** `test/AK-66663-infoblox-writeonly/`

- Create — All 4 WriteOnly fields show `(write-only attribute)`, credentials sent to API
- **KEY TEST:** State verification — `shared_secret` = `null`, `grid_master.username` = `null`, `grid_master.password` = `null`, `instance.password` = `null`
- API payload confirms `sharedSecret`, grid master `userName`/`password`, instance `password` correctly sent via `GetRawConfigAt` with indexed paths
- Import — Post-import plan shows no WriteOnly drift

### Brownfield Upgrade (Registry -> Fixed Build)

**Test environment:** `test/AK-66663-infoblox-writeonly/brownfield/`

- Create with registry — `shared_secret: "BfSharedSecret2024"` in plaintext state (the bug)
- Upgrade plan — No WriteOnly drift
- **KEY TEST:** `terraform refresh` scrubs plaintext credentials from state (values become `null`)
- Brownfield destroy — Clean

### Code Quality

- `make lint` — 0 issues
- `make build` — Clean
- `make doc` — Regenerated

### DIFF
--- a/alkira/resource_alkira_service_infoblox_helper.go
+++ b/alkira/resource_alkira_service_infoblox_helper.go
@@ -6,11 +6,50 @@ import (
 	"strconv"
 
 	"github.com/alkiranet/alkira-client-go/alkira"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+// getInfobloxWriteOnlyValue reads a WriteOnly field from raw config with a d.Get() fallback
+// for import/refresh where GetRawConfigAt is unavailable.
+func getInfobloxWriteOnlyValue(d *schema.ResourceData, field string) string {
+	attrPath := cty.Path{cty.GetAttrStep{Name: field}}
+	val, diags := d.GetRawConfigAt(attrPath)
+
+	if !diags.HasError() && !val.IsNull() && val.IsKnown() && val.Type() == cty.String {
+		if strVal := val.AsString(); strVal != "" {
+			return strVal
+		}
+	}
+
+	if v, ok := d.GetOk(field); ok {
+		return v.(string)
+	}
+
+	return ""
+}
+
+// getInfobloxNestedWriteOnlyValue reads a WriteOnly field from a nested TypeList block
+// via GetRawConfigAt with an indexed cty path.
+func getInfobloxNestedWriteOnlyValue(d *schema.ResourceData, block string, index int, field string) string {
+	attrPath := cty.Path{
+		cty.GetAttrStep{Name: block},
+		cty.IndexStep{Key: cty.NumberIntVal(int64(index))},
+		cty.GetAttrStep{Name: field},
+	}
+	val, diags := d.GetRawConfigAt(attrPath)
+
+	if !diags.HasError() && !val.IsNull() && val.IsKnown() && val.Type() == cty.String {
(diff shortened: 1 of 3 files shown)


alkira/resource_alkira_service_infoblox_helper.go
```diff
@@ -6,11 +6,50 @@ import (
...
+	attrPath := cty.Path{cty.GetAttrStep{Name: field}}
+	val, diags := d.GetRawConfigAt(attrPath)
+
+	if !diags.HasError() && !val.IsNull() && val.IsKnown() && val.Type() == cty.String {
+		if strVal := val.AsString(); strVal != "" {
+			return strVal
+		}
+	}
+
+	if v, ok := d.GetOk(field); ok {
```
snehachandrak: `d.GetOk()` will never return the actual value for a WriteOnly field — WriteOnly fields are null in state by definition. On import/refresh, `GetRawConfigAt` returns `null` (no config available) and `d.GetOk()` returns the zero value. So the fallback gives a false sense of safety while actually always returning "".

The `getInfobloxNestedWriteOnlyValue` function does NOT have this fallback, which is inconsistent. Either both helpers should have it or neither should. I'd recommend removing the fallback from `getInfobloxWriteOnlyValue` to match `getInfobloxNestedWriteOnlyValue` and avoid confusion — the caller already handles the ""  case correctly (the credential is only created when `shared_secret != ""`).

PR #624, opened 2026-04-18
AK-67141: fix authentication_mode not populated during import due to …

yashgupta912121:
**Bug**: authentication_mode was not populated during import because d.Set() received a []string (from API's SupportedModes) on a TypeString schema field. Terraform silently ignores this type mismatch, so the field was never written to state. This caused terraform plan -generate-config-out to fail with "Missing required argument".         
   
  Fix: Extract SupportedModes[0] before calling d.Set().                                                              
                                                                                                                   
  Tested on yash tenant (yash.quarterbacks.dev.alkira2.net) against existing connector                                                                                      
                                                                                                                      
  - Ran terraform plan -generate-config-out=generated.tf using the fixed provider binary (branch task/AK-67141)       
  - **Before fix:** generated config has no value for authentication_mode, plan fails with Error: Missing required argument                                                                                                            
  - **After fix:** generated config correctly emits authentication_mode = "LOCAL", plan succeeds with 1 to import, 0 to 
  add, 0 to change, 0 to destroy                                                                                      
  - subnet field also correctly populated as "10.0.0.0/24" (string, not array)

### DIFF
--- a/alkira/resource_alkira_connector_remote_access_helper.go
+++ b/alkira/resource_alkira_connector_remote_access_helper.go
@@ -200,7 +200,9 @@ func setAuthorization(d *schema.ResourceData, segmentOptions []alkira.ConnectorR
 // setConnectorRemoteAccess
 func setConnectorRemoteAccess(connector *alkira.ConnectorRemoteAccessTemplate, d *schema.ResourceData, m interface{}) error {
 
-	d.Set("authentication_mode", connector.AuthenticationOptions.SupportedModes)
+	if len(connector.AuthenticationOptions.SupportedModes) > 0 {
+		d.Set("authentication_mode", connector.AuthenticationOptions.SupportedModes[0])
+	}
 
 	// Set ldap_settings block
 	if connector.AuthenticationOptions.LdapSettings != nil {
--- a/alkira/resource_alkira_connector_remote_access_helper.go
+++ b/alkira/resource_alkira_connector_remote_access_helper.go
@@ -200,6 +200,11 @@ func setAuthorization(d *schema.ResourceData, segmentOptions []alkira.ConnectorR
 // setConnectorRemoteAccess
 func setConnectorRemoteAccess(connector *alkira.ConnectorRemoteAccessTemplate, d *schema.ResourceData, m interface{}) error {
 
+	if len(connector.AuthenticationOptions.SupportedModes) > 1 {
+		log.Printf("[WARN] connector-remote-access %q has %d authentication modes; only the first (%q) will be managed",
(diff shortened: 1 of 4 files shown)


asim-ak (changes requested):
The fix is right -- `d.Set()` silently drops the value when you pass `[]string` to a `TypeString` field, so extracting `[0]` is the correct approach.

A few things worth addressing:

### `Arguments[0]` can panic (pre-existing, same function)

Lines 224-226 hit `connector.Arguments[0]` without checking length. Empty array = panic = Terraform crash. Same class of bug you're fixing here, and you're already in this function:

```go
if len(connector.Arguments) == 0 {
    return fmt.Errorf("API returned empty arguments for connector-remote-access %q", connector.Name)
}
```

### Multiple modes get silently dropped

Backend model is `List<Mode>` with no max-length constraint. If someone creates a connector through the portal with `["LOCAL", "LDAP"]`, importing it takes only `[0]` and the next apply sends `["LOCAL"]` back -- quietly removing LDAP. A warning log would help:

```go
if len(connector.AuthenticationOptions.SupportedModes) > 1 {
    log.Printf("[WARN] connector-remote-access %q has %d authentication modes; only the first (%q) will be managed",
        connector.Name, len(connector.AuthenticationOptions.SupportedModes),
        connector.AuthenticationOptions.SupportedModes[0])
}
```

### No tests for this resource

There's no test file for `connector_remote_access` at all. Basic coverage for `setConnectorRemoteAccess` -- happy path, empty modes, empty arguments -- would go a long way.

### Summary

| Finding | Severity | Required? |
|---------|----------|-----------|
| `Arguments[0]` panic | High (pre-existing) | Recommended |
| Multi-mode truncation | Medium | Nice to have |
| No unit tests | Medium | Recommended |

The fix itself is correct. Main ask is the `Arguments[0]` guard since you're already in this function.

yashgupta912121:
added the fix as per the comment, did this additional set of testing

  1. Create — Connector 276 created with authentication_mode = "LOCAL"                                                
  2. Import + generate-config-out — generated.tf correctly shows authentication_mode = "LOCAL" (line 6). Before the
  fix, this field would be missing and plan would fail.                                                               
  3. Plan on existing resource — No diff on authentication_mode. The only diff shown is the authorization block's  
  billing_tag_id/prefix_list_id defaults (null vs 0) which is a pre-existing cosmetic issue unrelated to this bug.    
                                                                                                                   
  The fix is verified. SupportedModes[0] correctly populates authentication_mode in state during Read.

yashgupta912121:
We can add UT for this file as a Add on PR

asim-ak (approved):

PR #627, opened 2026-04-23
AK-67216 fix spurious PROVISION (DELETE) FAILED warning on policy res…

camjrichards:
…ources

Gate the DELETE diagnostic on provErr != nil (matching CREATE/UPDATE) instead of provState != "SUCCESS". The SDK's delete() legitimately returns provState="" with nil err/valErr/provErr on 202 async-validation, 404 already-deleted, and provision-disabled paths; the old guard fired on all three and printed a warning whose Detail was %!s(<nil>), which ATF's stdout matcher read as a test failure.

Applies to alkira_policy, alkira_policy_rule, and alkira_policy_rule_list.

### DIFF
--- a/alkira/resource_alkira_policy.go
+++ b/alkira/resource_alkira_policy.go
@@ -225,7 +225,7 @@ func resourcePolicyDelete(ctx context.Context, d *schema.ResourceData, m interfa
 	client := m.(*alkira.AlkiraClient)
 	api := alkira.NewTrafficPolicy(m.(*alkira.AlkiraClient))
 
-	provState, err, valErr, provErr := api.Delete(d.Id())
+	_, err, valErr, provErr := api.Delete(d.Id())
 
 	if err != nil {
 		// Terraform may not print "with <resource address>" for destroys of objects
@@ -248,7 +248,7 @@ func resourcePolicyDelete(ctx context.Context, d *schema.ResourceData, m interfa
 
 	d.SetId("")
 
-	if client.Provision && provState != "SUCCESS" {
+	if client.Provision && provErr != nil {
 		return diag.Diagnostics{{
 			Severity: diag.Warning,
 			Summary:  "PROVISION (DELETE) FAILED",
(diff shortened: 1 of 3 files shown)


moiz-alkira (approved):
1. Same bug exists in ~40 other resource files. Every other resource_alkira_*.go that has a delete path uses the identical provState != "SUCCESS" guard. A non-exhaustive list:
  connector_aws_vpc, connector_azure_vnet, connector_ipsec, service_fortinet, service_cisco_ftdv, service_f5_lb, internet_applications, byoip_prefix, ip_reservation, flow_collector,
  segment_resource_share, policy_routing, policy_nat_rule, list_* (all of them), etc. Any ATF test that destroys one of those resources on a 202/404/provision-disabled path will hit the same
   spurious-warning failure. Strongly recommend either (a) expanding this PR to fix all of them, or (b) filing a follow-up ticket explicitly tracking the broader sweep — otherwise this fix
  will read like whack-a-mole the next time ATF lights up.
  2. Detail-field nil rendering is the underlying anti-pattern. Even after this fix, several other diagnostic sites in the codebase still do Detail: fmt.Sprintf("%s", someErr) without a nil
  guard. If any of them ever fire with a nil error, you'll get %!s(<nil>) again. Worth a defensive helper like errDetail(e error) string { if e == nil { return "" }; return e.Error() } as a
  future cleanup — out of scope here, but worth noting.
  3. No test added. The PR doesn't include a unit/acceptance test that pins the 202/404/provision-disabled behavior. Given this exact bug class will likely recur (see #1), a small
  table-driven test against the diagnostic emission would prevent regressions. Acceptable to defer if the ATF run that originally caught it is now treated as the regression gate — but worth
  saying so in the PR description.

snehachandrak:
@camjrichards add task under [https://alkiranet.atlassian.net/browse/AK-67234](url) - to track this for rest of the resources.

snehachandrak (approved):

PR #664, opened 2026-05-13
[AK-68129] Adds Computed to customer_asn schema for Azure VNet connec…

jyoti-r27:
…tor to stop perpetual plan diff in VGW mode

## Root cause

  The provider declares `customer_asn` as `Optional` only, but the backend auto-populates the field when the user omits it in VGW mode — either to the existing Azure VGW's ASN
  (`AzureVNETConnectorServiceImpl.java:504`) or, when no VGW pre-exists, to `DEFAULT_ASN = 64521` (`AzureVNETConnectorServiceImpl.java:151,1421,1669`). The Read function writes that
  backend-supplied value into Terraform state, and because the schema treats the field as user-owned, the populated state reads as drift versus the empty config — producing a perpetual
  "remove `customer_asn`" diff.

  Apply doesn't resolve it because every leg of the cycle treats "field absent" as "don't touch": the provider builds the update payload from `d.Get("customer_asn").(int) = 0`, the Go
  client's `omitempty` drops `0` from the JSON, the backend update path leaves the stored value alone when the field is absent, and Read writes the persisted value back into state.

  ## Fix

  Mark `customer_asn` as `Optional + Computed` in the schema. This tells Terraform that the field can be set by the user *or* by the backend, and a backend-supplied value is a valid
  settled outcome — no drift.

  ```go
  "customer_asn": {
      Type:     schema.TypeInt,
      Optional: true,
      Computed: true,   // <-- added
  },

  Users who explicitly set customer_asn are unaffected: the backend already rejects any user-supplied ASN that doesn't match the VGW (AzureVNETConnectorServiceImpl.java:501-502), so the
  read-back value matches the config.

  Backend changes are not needed — the auto-set behavior is correct (BGP requires a customer ASN; either the VGW's existing ASN or DEFAULT_ASN is the right product choice). The bug was
  only that the provider schema didn't reflect that the field is co-owned.

### DIFF
--- a/alkira/resource_alkira_connector_azure_vnet.go
+++ b/alkira/resource_alkira_connector_azure_vnet.go
@@ -261,9 +261,12 @@ func resourceAlkiraConnectorAzureVnet() *schema.Resource {
 					"when `connection_mode` is `VNET_PEERING`. This field cannot be updated " +
 					"once the connector has been provisioned. The ASN cannot be value that " +
 					"is [restricted by Azure]" +
-					"(https://learn.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-vpn-faq#bgp).",
+					"(https://learn.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-vpn-faq#bgp). " +
+					"If omitted, the backend assigns one (the existing Azure VGW's ASN if " +
+					"present, otherwise a default), which the provider reads back into state.",
 				Type:     schema.TypeInt,
 				Optional: true,
+				Computed: true,
 			},
 			"scale_group_id": {
 				Description: "The ID of the scale group associated with the connector.",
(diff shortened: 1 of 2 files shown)


moiz-alkira:
Nice diagnosis and clean fix for the perpetual-diff case 👍

One thing I wanted to check — did you look into `DiffSuppressFunc` for handling post-create edits to `customer_asn`?

With the current change, `Optional + Computed` resolves the "user omits the field" path, but if a user later *changes* `customer_asn` on an existing connector, Terraform will still call `Update`, the backend will ignore it (since the field is immutable post-provisioning per the description), and the next refresh will show drift again — just with a different shape.

`DiffSuppressFunc` would make the field create-time-only without destroying the resource or erroring out:

```go
"customer_asn": {
    Type:     schema.TypeInt,
    Optional: true,
    Computed: true,
    DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
        // Field is immutable post-create; ignore config changes on existing resources.
        return d.Id() != ""
    },
},
```

On create, `d.Id()` is empty and the user's value is honored. On any subsequent plan, edits to `customer_asn` are suppressed as no-ops, so state and config stay aligned without replacement or error.

Happy to keep it out of scope for this PR since the perpetual-diff fix stands on its own — just wanted to flag it in case it's worth a follow-up.

moiz-alkira (approved):

jyoti-r27:
Ack, will address this in follow-up PR.

PR #670, opened 2026-05-13
AK-67730: Fix Checkpoint/FTDv management server perpetual diff

snehachandrak:
Problem:
Commit 9a131d9a (AK-63243) corrected the deflate field names for Checkpoint management_server and Cisco FTDv firepower_management_center ("segment" -> "segment_id", "user_name" -> "username"). Since both blocks use TypeSet, the corrected field names change the hash on the first Read after upgrade from v1.4.4. Terraform sees different hashes as "old block removed, new block added" (remove+add). The "add" side is built from HCL config which has no credential_id (it's Computed), so the expand function creates a new credential on every apply. This cycle never stabilizes — perpetual diff with orphaned credentials.

Additionally, the API doesn't return sensitive fields (password for Checkpoint, username/password for FTDv). Without carrying these from prior state, every Read writes empty values causing a perpetual diff on those fields.

Fix:
1. Change management_server (Checkpoint) and firepower_management_center (Cisco FTDv) from TypeSet to TypeList with MaxItems=1. TypeList compares fields directly instead of hashing, so the corrected field names don't trigger remove+add, and credential_id is preserved from state during in-place updates.

2. Add d.GetOk pattern to both deflate functions to carry forward password (Checkpoint) and username/password (FTDv) from prior state, since the API doesn't return these sensitive fields.

Changes per resource:
- Schema: TypeSet -> TypeList + MaxItems: 1
- Expand call site: .(*schema.Set) -> .([]interface{})
- Expand function: *schema.Set -> []interface{}
- Deflate function: accepts *schema.ResourceData, carries sensitive fields from prior state via d.GetOk

Verified:
- Created Checkpoint + FTDv with v1.4.4, upgraded to this fix
- management_server perpetual diff: FIXED
- firepower_management_center perpetual diff: FIXED
- credential_id cycling: FIXED (no orphaned credentials)
- password/username perpetual diff: FIXED (carried from state)
- Edited Checkpoint management_server.ips: apply succeeds, plan clean
- Edited FTDv firepower_management_center.server_ip: apply succeeds, plan clean
- Multiple apply cycles: no perpetual diff on either resource
- Only remaining diff: Checkpoint segment_options DEFAULT zone (pre-existing in v1.4.4)
- Tested on sneha.grassroots.dev.alkira2.net (CXP: US-WEST)

### DIFF
--- a/alkira/resource_alkira_service_checkpoint.go
+++ b/alkira/resource_alkira_service_checkpoint.go
@@ -122,7 +122,8 @@ func resourceAlkiraCheckpoint() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"BRING_YOUR_OWN", "PAY_AS_YOU_GO"}, false),
 			},
 			"management_server": {
-				Type:        schema.TypeSet,
+				Type:        schema.TypeList,
+				MaxItems:    1,
 				Required:    true,
 				Description: "",
 				Elem: &schema.Resource{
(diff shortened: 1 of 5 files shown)


sankeerthmogili (approved):
Looks good. TypeSet→TypeList+MaxItems=1 is the right fix for the hash-based perpetual diff, and the d.GetOk carry-forward pattern for sensitive fields is correct.

One minor follow-up to consider: verify that the Checkpoint API actually returns `username` — if not, it should be carried forward from state like `password` is. The FTDv side already carries forward both.

snehachandrak:
> h.

  - Checkpoint: API returns username but not password → only password needs d.GetOk carry-forward
  - FTDv: API returns neither username nor password → both need d.GetOk carry-forward

PR #673, opened 2026-05-14
AK-66660: Mark PAN service credential fields sensitive and add plan-time validation

asim-ak:
## Summary

- Mark 7 credential fields on `alkira_service_pan` as `Sensitive: true` (`pan_password`, `pan_license_key`, `master_key`, `registration_pin_id`, `registration_pin_value`, instance.`auth_key`, instance.`auth_code`)
- Add `ValidateFunc` on `pan_username` restricting values to `admin` or `akadmin`
- Add `CustomizeDiff` rules for per-CXP `pan_username` enforcement, required `master_key` when `master_key_enabled=true`, and instance `name` uniqueness
- Emit a `Warning` diagnostic on Update when bootstrap-only credential fields change
- Fix 4 pre-existing helper bugs

## Problem

Credential values entered in HCL appeared in plaintext in `terraform plan`, `apply`, and `show` output, and in state-file diffs. The backend already enforced a per-CXP rule (Azure: akadmin, AWS/GCP: admin) and a username regex (`^(admin|akadmin)$`), but only at apply time. Multi-instance configurations with duplicate or empty `name` values silently corrupted state, because `setPanInstances` matched by name when sourcing bootstrap-only `auth_*` fields from prior state.

Pre-existing helper bugs:
- `createPanMasterKeyCredential` logged `master_key_expiry` before checking `master_key_enabled`. The empty-expiry check ran after `convertInputTimeToEpoch`, so users hit a confusing time-parse error instead of a clear required-field error.
- `updatePanCredential` used `nil` checks on TypeString fields. `Get` returns `""`, not `nil`, so those branches never triggered.
- `setPanInstances` matching loop did not `break`, so later instances overwrote earlier `authKey`, `authCode`, and `authExpiry` values.
- Comment typo `credentail` in two places.

### Sensitive fields
Added `Sensitive: true` to the 7 fields listed above. `pan_username` is non-sensitive because the value is constrained to `admin` or `akadmin` and is not a secret. Date fields (`master_key_expiry`, `registration_pin_expiry`, `auth_expiry`) are also non-sensitive.

### Plan-time validation
- `ValidateFunc: validation.StringInSlice([]string{"admin", "akadmin"}, false)` on `pan_username`
- `CustomizeDiff` switch on `cxp.Provider` that rejects username/CXP mismatches at plan time
- `CustomizeDiff` gate on `master_key_enabled` requiring `master_key` and `master_key_expiry` on Create and on false→true Update transitions. The gate is skipped on idempotent refresh after import, because the API does not return `master_key_expiry`.
- `CustomizeDiff` rule rejecting duplicate `name` on multi-instance configurations. Empty `name` is rejected on Create only, so legacy state with empty names is tolerated on Update.

### Bootstrap-only warning
`resourceServicePanUpdate` emits a `Warning` when `master_key`, `master_key_expiry`, `registration_pin_*`, or per-instance `auth_*` fields change. The backend's `PanBootstrapTaskHandler` consumes these only on `CREATE`, so there is no in-place rotation flow. Changes are recorded in state but never reach the PAN VM.

### Helper bug fixes
- `createPanMasterKeyCredential` now short-circuits on `master_key_enabled=false` before any logging. The empty-expiry check runs before `convertInputTimeToEpoch` and returns a clear error.
- `updatePanCredential` `nil` checks changed to `== ""`. Flattened the `else` branch.
- `setPanInstances` adds a `break` after the first match. The matcher prefers a non-zero `id`, then falls back to a non-empty `name`. CustomizeDiff already enforces name uniqueness, so the fallback is unambiguous.
- New helper `panInstanceAuthFieldsChanged` does an index-based comparison and is used by the Update warning.

## Changes

- `alkira/resource_alkira_service_pan.go` — schema Sensitive flags, ValidateFunc, CustomizeDiff (per-CXP rule, master_key gate, instance name uniqueness), Update warnings
- `alkira/resource_alkira_service_pan_helper.go` — 4 bug fixes, new `panInstanceAuthFieldsChanged` helper, tightened `setPanInstances` matcher
- `alkira/resource_alkira_service_pan_test.go` — unit tests for Sensitive flags, ValidateFunc, helper bug fixes, matcher preservation, Update-warning helper (+243 lines)
- `templates/resources/service_pan.md.tmpl` — new doc template covering credential types, update semantics, import behavior, and state-file warning
- `docs/resources/service_pan.md` — regenerated
- `examples/resources/alkira_service_pan/resource.tf` — rewritten to use `variable` blocks with `sensitive = true` for credentials. The previous version had hardcoded values that would have published to the Terraform Registry.
- `docs/changes/{README.md, unreleased.md}` — new changelog convention plus the AK-66660 entry

### Live Testing

**Test environment:** `test/AK-66660/` (15 scenarios across 6 subdirectories)

- S1 Greenfield Create, Read, idempotent Plan. All 7 sensitive fields render as `(sensitive value)` in `show`. State file contains plaintext as documented. Post-apply plan shows No changes.
- S2 ValidateFunc rejection. `pan_username = "root"`, `""`, `"Admin"` (case-sensitive) all rejected at plan time.
- S3 Per-CXP rule. AWS CXP rejects `akadmin` with the new error message. `admin` passes. Azure CXP rejects `admin` as a side-test.
- S4a Master-key gate on Create. Empty `master_key` and empty `master_key_expiry` rejected separately.
- **KEY TEST:** S4b Post-create plan returns No changes. The master-key gate does not re-fire on idempotent refresh, which is the false-positive concern that drove the `isCreate || isEnabling` design.
- S4c Master-key gate on Update. false→true transition with empty `master_key` rejected.
- S5 Bogus CXP. Plan succeeds with `[WARNING] Unable to get CXP information` logged. No plan-time false positive on transient API failure.
- **KEY TEST:** S6 `auth_key` preserved across `terraform refresh`. instance.id is non-zero, confirming the id-match path. Regression guard for the `setPanInstances` matcher rewrite.
- S7 Duplicate or empty name. Both rejected at plan time with explicit error.
- S8 Bootstrap-only Warning fires on `master_key` change. One warning per apply.
- S9 Idempotent re-apply emits 0 bootstrap warnings.
- **KEY TEST:** S10 `pan_password` rotation. `credential_id` is unchanged across update. `PUT /credentials/pan/{id}` issued. 0 bootstrap warnings. Vault is rotated in place, no orphan credential.
- S11 Per-instance `auth_key` change emits the "PAN instance auth fields are bootstrap-only" Warning.
- S12 Destroy. `DELETE /panfwservices/{id}` issued and state is cleared.

Full results in `test/AK-66660/TEST_PLAN.md`.

### Unit Tests

- `TestAlkiraServicePanResourceSchema` asserts Sensitive flags on the 5 top-level and 2 instance fields, asserts `Sensitive: false` on `pan_username` and `auth_expiry`, and asserts `ValidateFunc` non-nil
- `TestAlkiraServicePanValidateUsername` — 5 cases (admin, akadmin, root, "", Admin)
- `TestAlkiraServicePanInstanceAuthFieldsChanged` — 4 cases (no-change, auth_key changed, instance added, non-auth changed)
- `TestAlkiraServicePanSetPanInstancesPreservesAuthFields` — 3 sub-tests (post-create name-match, post-refresh id-match, no-prior-state empty)
- `TestAlkiraServicePanCreatePanMasterKeyCredential` — disabled short-circuit, empty-expiry rejection
- `TestAlkiraServicePanUpdatePanCredential` — no-op early-return

### Code Quality

- `make lint` — 0 issues
- `go build` — clean
- `go test ./alkira/...` — all pass (105s)


snehachandrak (changes requested):
needs a rebase but clean PR. 
All validations align with TPS. No drift risk. Backward compatible for valid existing configs.

PR #676, opened 2026-05-18
AK-68322: Add loopback_prefixes to alkira_connector_aws_dx

shrey-pasari:


alkira/resource_alkira_connector_aws_dx.go
```diff
@@ -96,6 +96,15 @@ func resourceAlkiraConnectorAwsDx() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"loopback_prefixes": {
+				Description: "A list of prefixes used to allocate loopback " +
+					"IPs across DX instances. Required when scaling DX " +
```
amitmandke: when using tunnel scale options right? I think it is iportant to mention that as only the scale options needs this. not instances. without scale options every instance accepts required loopbacks correctly.

alkira/resource_alkira_connector_aws_dx.go
```diff
@@ -96,6 +96,15 @@ func resourceAlkiraConnectorAwsDx() *schema.Resource {
...
+					"horizontally with multiple instances. Eg: " +
```
amitmandke: Also we should mentioned these are /26 prefixes provided by Alkira.

alkira/resource_alkira_connector_aws_dx.go
shrey-pasari: Done

alkira/resource_alkira_connector_aws_dx.go
shrey-pasari: Done

amitmandke (approved):

PR #683, opened 2026-05-19
AK-66965: Add scale_group_id to alkira_connector_azure_vnet_third_party

shrey-pasari:
## Summary
  - Add `scale_group_id` (Optional, String) to the `alkira_connector_azure_vnet_third_party` schema
  - Wire it through the Read path and the request generator so users can attach the connector to a scale group
  - Bump vendored `alkira-client-go` to a version that exposes `ScaleGroupId` on `AzureVnetThirdPartyConnector`
  - Add unit tests covering JSON marshal omission when unset, inclusion when set, round-trip, and schema shape (Optional/TypeString)

  ## Notable decisions
  - Field is `Optional` (not `Required` or `Computed`) — matches `scale_group_id` conventions on other connectors (e.g. AWS DX, GCP Interconnect)
  - SDK struct uses `json:"scaleGroupId,omitempty"` so existing tfstates without a scale group don't start sending an empty value to the backend
  - Tests assert `scaleGroupId` is *omitted* from the JSON body when empty, to guard against future struct-tag regressions

  ## Test plan
  - [ ] `make lint` clean
  - [ ] `go test ./alkira/ -run TestConnectorAzureVnetThirdPartyScaleGroupId -run TestAzureVnetThirdPartyScaleGroupIdSchema -run TestAzureVnetThirdPartyScaleGroupIdRequest`
  - [ ] Apply a tf config that sets `scale_group_id` and confirm it is sent in the create request and round-trips on refresh
  - [ ] Apply a tf config without `scale_group_id` and confirm no `scaleGroupId` is sent

### DIFF
--- a/alkira/resource_alkira_connector_azure_vnet_third_party.go
+++ b/alkira/resource_alkira_connector_azure_vnet_third_party.go
@@ -69,6 +69,17 @@ func resourceAlkiraConnectorAzureVnetThirdParty() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 			},
+			"scale_group_id": {
+				Description: "The ID of a scale group to place this " +
+					"connector on. A scale group is a set of " +
+					"dedicated CXP nodes used to isolate a connector " +
+					"or distribute many connectors for performance. " +
+					"Typically available only on high-capacity CXPs " +
+					"(10Gb and above). Can only be set at create " +
+					"time and cannot be changed after provisioning.",
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 			"azure_vnet_third_party_connector_attachment_id": {
 				Description: "The ID of the Azure VNET Third Party Connector Attachment.",
 				Type:        schema.TypeInt,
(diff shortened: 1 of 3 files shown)


alkira/resource_alkira_connector_azure_vnet_third_party.go
```diff
@@ -69,6 +69,11 @@ func resourceAlkiraConnectorAzureVnetThirdParty() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 			},
+			"scale_group_id": {
+				Description: "The ID of the scale group associated with the connector.",
```
amitmandke: its not associated with a connector, its a choice that customer makes for a connector. Scale group represents additional resources (set of nodes) on CXP. Once made available a scale group id can be selected to isolate required connector(s) on these dedicated resources or distribute high number of connectors across these additional resources to get desired performance out of these connectors. This is usually needed (and/or made available) only on hight capacity CXPs (10Gb and above). This can be selected only before provisioning the connector once provisioned a scale group cannot be edited.

shrey-pasari: Done

amitmandke (approved):

amitmandke (approved):

PR #684, opened 2026-05-20
AK-68326: add SCM + Advanced Routing to alkira_service_pan

ramakanthg:
## Summary
- New optional + ForceNew schema attrs on `alkira_service_pan`: `scm_enabled`, `scm_folder`, `routing_type`. ForceNew matches TPS update-path immutability.
- `CustomizeDiff` enforces the panorama/scm mutex and `routing_type=advanced` implies `scm_enabled` at plan time.
- New example `examples/resources/alkira_service_pan/resource_scm.tf` for SCM mode; existing Panorama example untouched.
- Manual SCM section added to `docs/resources/service_pan.md` above the auto-generated schema block.

## Context
Part of [AK-67476](https://alkiranet.atlassian.net/browse/AK-67476) (PAN FW SCM + Advanced Routing).

## Blocking on
**alkiranet/alkira-client-go#249** — must merge + tag, then bump dep here. Local build verified with a `replace` directive (not in this commit). DRAFT until then.

## Test plan
- [ ] Bump `alkira-client-go` dep once #249 is tagged; confirm CI green.
- [ ] `terraform plan` on `resource_scm.tf` example.
- [ ] Verify panorama/scm mutex error surfaces at plan time.
- [ ] Verify `routing_type=advanced` without `scm_enabled=true` errors at plan time.

[AK-67476]: https://alkiranet.atlassian.net/browse/AK-67476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alkira/resource_alkira_service_pan.go
```diff
@@ -48,6 +48,17 @@ func resourceAlkiraServicePan() *schema.Resource {
...
 			}
 
+			panoramaEnabled := d.Get("panorama_enabled").(bool)
+			scmEnabled := d.Get("scm_enabled").(bool)
+			routingType := d.Get("routing_type").(string)
+
+			if panoramaEnabled && scmEnabled {
+				return fmt.Errorf("[ERROR] panorama_enabled and scm_enabled are mutually exclusive")
+			}
+			if routingType == "advanced" && !scmEnabled {
```
snehachandrak, line 58: Need to add the reverse validation. A user who sets scm_enabled = true without routing_type = "advanced" will pass plan but get a 400 on apply. Add:

  ```
if scmEnabled && routingType != "advanced" {
      return fmt.Errorf("[ERROR] scm_enabled=true requires routing_type=\"advanced\"")
  }
```

File: alkira/resource_alkira_service_pan.go [hunk 2]
```diff
@@ -298,6 +309,39 @@ func resourceAlkiraServicePan() *schema.Resource {
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+			},
+			"scm_folder": {
+				Description: "SCM folder path for this firewall (e.g. " +
+					"`prod` or `prod/edge/site-a`). 1-4 segments separated " +
+					"by `/`; each segment may contain alphanumeric, hyphen, " +
+					"underscore, period, and space. Max total length 64. Only " +
+					"valid when `scm_enabled = true`. Immutable once the " +
```
snehachandrak, line 333: The description says "Only valid when scm_enabled = true", and TPS rejects it when SCM is off. Catch it at plan time by adding the below to CustomizeDiff:

```
  if !scmEnabled && d.Get("scm_folder").(string) != "" {
      return fmt.Errorf("[ERROR] scm_folder requires scm_enabled=true")
  }
```

alkira/resource_alkira_service_pan_helper.go
```diff
@@ -430,6 +430,8 @@ func generateServicePanRequest(d *schema.ResourceData, m interface{}) (*alkira.S
 	panoramaDeviceGroup := d.Get("panorama_device_group").(string)
 	panoramaIpAddresses := convertTypeListToStringList(d.Get("panorama_ip_addresses").([]interface{}))
 	panoramaTemplate := d.Get("panorama_template").(string)
+	scmFolder := d.Get("scm_folder").(string)
```
snehachandrak: **Blocking bug:** When the user doesn't set `scm_folder`, this sends `"scmFolder": ""` to TPS. TPS has a `@Pattern` regex on that field that requires at least one character, so `""` fails validation and returns a 400. `null` would be fine (the regex skips null), but this code always sends a non-null empty string.

This breaks every PAN service create/update that doesn't set `scm_folder`.

**Fix:** Remove these two lines, remove `RoutingType: &routingType` and `ScmFolder: &scmFolder` from the struct literal (lines 483-485), and add after the struct:

```go
if v, ok := d.GetOk("scm_folder"); ok {
    s := v.(string)
    request.ScmFolder = &s
}
if v, ok := d.GetOk("routing_type"); ok {
    s := v.(string)
    request.RoutingType = &s
}
```

This way unset fields stay `nil` and get omitted from JSON entirely.

alkira/resource_alkira_service_pan.go
ramakanthg: Added the missing SCM→AR direction in CustomizeDiff — `scm_enabled=true` without `routing_type="advanced"` now errors at plan time. Together with the existing AR→SCM check this forms a biconditional (SCM ⇔ AR): you must have both or neither, never just one. (Earlier I called this a "reverse mutex" — wrong term; it is the reverse *implication* completing a biconditional.) Fixed in 5e5af16f.

alkira/resource_alkira_service_pan.go
ramakanthg: Added the plan-time guard in CustomizeDiff — `scm_folder` set without `scm_enabled=true` now errors before apply. Fixed in 5e5af16f.

alkira/resource_alkira_service_pan_helper.go
ramakanthg: Good catch — switched both `ScmFolder` and `RoutingType` to `d.GetOk` so unset fields stay `nil` and get omitted from the JSON via `omitempty`. Fixed in 5e5af16f.

alkira/resource_alkira_service_pan.go
```diff
@@ -553,6 +603,14 @@ func resourceServicePanRead(ctx context.Context, d *schema.ResourceData, m inter
 		d.Set("panorama_template", pan.PanoramaTemplate)
 	}
 
+	d.Set("scm_enabled", pan.ScmEnabled)
```
snehachandrak, line 606: you can run the following to update the client.

```
go get github.com/alkiranet/alkira-client-go@main
go mod tidy
go mod vendor
```

ramakanthg: @snehachandrak - merged the client and update this PR, pls check now.

snehachandrak (approved):

PR #689, opened 2026-05-21
AK-68496: [CLONE-65] Add loopback_prefixes to alkira_connector_aws_dx

shrey-pasari:


vendor/github.com/alkiranet/alkira-client-go/alkira/connector_aws_direct_connect.go
```diff
@@ -40,18 +40,19 @@ type ConnectorAwsDirectConnectInstance struct {
...
-	ImplicitGroupId int                                 `json:"implicitGroupId,omitempty"` // response only
-	Size            string                              `json:"size"`
-	ScaleGroupId    string                              `json:"scaleGroupId,omitempty"`
-	TunnelProtocol  string                              `json:"tunnelProtocol"`
-	Instances       []ConnectorAwsDirectConnectInstance `json:"instances"`
+	Id               json.Number                         `json:"id,omitempty"` // response only
+	Name             string                              `json:"name"`
+	Description      string                              `json:"description,omitempty"`
+	BillingTags      []int                               `json:"billingTags"`
+	Cxp              string                              `json:"cxp"`
```
sankeerthmogili, line 47: This is missing `omitempty` — when a user doesn't set `loopback_prefixes`, this will serialize as `"loopbackPrefixes": null` in the API request. On the TPS side, the update path (`DirectConnectConnectorServiceImpl`) unconditionally overwrites the stored value with whatever the request contains, so this would null out existing loopback prefixes on connectors that already have them set.

There's a guard when scale options exist (TPS returns 400), but without scale options the null goes through silently. We should fix both sides — `omitempty` here, and a TPS-side fix to preserve existing values when null is sent on update. Can you add `omitempty` to the JSON tag?

sankeerthmogili: cc: @amitmandke

shrey-pasari: updated the alkira-client and updated the terraform with the client

amitmandke (approved):

PR #693, opened 2026-05-29
AK-67764: Add policy_fqdn_list_id and INTERNAL_DNS to internet applic…

camjrichards:
…ation resource

Add `policy_fqdn_list_id` as an optional `TypeInt` attribute on the target block and accept `INTERNAL_DNS` as a valid target type. Field is plumbed through expand/flatten so create, read, and import round-trip correctly. Existing `IP`/`ILB_NAME` configurations are unaffected.

### DIFF
--- a/alkira/resource_alkira_internet_applications.go
+++ b/alkira/resource_alkira_internet_applications.go
@@ -183,16 +183,16 @@ func resourceAlkiraInternetApplication() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"type": {
 							Description: "The type of the target, one of " +
-								"`IP` or `ILB_NAME`.",
+								"`IP`, `ILB_NAME` or `INTERNAL_DNS`.",
 							Type:     schema.TypeString,
 							Required: true,
 							ValidateFunc: validation.StringInSlice(
-								[]string{"IP", "ILB_NAME"}, false),
+								[]string{"IP", "ILB_NAME", "INTERNAL_DNS"}, false),
 						},
 						"value": {
-							Description: "IFA ILB name or private IP.",
+							Description: "IFA ILB name or private IP. Not required when `type` is `INTERNAL_DNS`.",
 							Type:        schema.TypeString,
-							Required:    true,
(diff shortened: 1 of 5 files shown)


alkira/resource_alkira_internet_applications.go
```diff
@@ -183,16 +183,16 @@ func resourceAlkiraInternetApplication() *schema.Resource {
...
 							ValidateFunc: validation.StringInSlice(
-								[]string{"IP", "ILB_NAME"}, false),
+								[]string{"IP", "ILB_NAME", "INTERNAL_DNS"}, false),
 						},
 						"value": {
-							Description: "IFA ILB name or private IP.",
+							Description: "IFA ILB name or private IP. Not required when `type` is `INTERNAL_DNS`.",
 							Type:        schema.TypeString,
-							Required:    true,
+							Optional:    true,
```
moiz-alkira, line 248: When we have an optional "required", the ideal way to implement is using a Terraform validator instead of simply relying on the backend to do the validation. The reason is that you want to catch the issue when doing a terraform plan vs terraform apply. Consider adding a validator.

camjrichards: Updated the `CustomizeDiff` function for internet applications to validate that value is present for IP and ILB NAME, and that policy_fqdn_list_id is present for INTERNAL_DNS

sankeerthmogili (commented):
Docs need regenerating — `docs/resources/internet_application.md` still shows `value` as Required and is missing `policy_fqdn_list_id`. Run `go generate ./...` to pick up the updated schema descriptions. Consider adding an INTERNAL_DNS example too.

camjrichards:
> Docs need regenerating — `docs/resources/internet_application.md` still shows `value` as Required and is missing `policy_fqdn_list_id`. Run `go generate ./...` to pick up the updated schema descriptions. Consider adding an INTERNAL_DNS example too.

Update the docs to add the updated descriptions

sankeerthmogili (approved):

PR #701, opened 2026-06-05
AK-69357: warn when config changes are skipped on FAILED resources

camjrichards:
## Summary

- Add a `warnOnFailedStateUpdate` UpdateContext wrapper that emits a non-fatal `CONFIGURATION CHANGES SKIPPED` warning when an apply carrying config changes targets a resource in `FAILED` provision state
- Wire the wrapper into all 55 resources that use the `CustomizeDiff` FAILED→SUCCESS retry pattern
- Document the retry-on-FAILED / skip-modify behavior in the PROVISIONING section of the provider docs (`templates/index.md.tmpl` + `docs/index.md`)

## Root cause ([AK-69357](https://alkiranet.atlassian.net/browse/AK-69357))

When a connector is in `FAILED` state and provision mode is enabled, the provider re-sends a PUT with `?provision=true` on re-apply — the supported retry-provisioning mechanism. TPS (`AbstractConnectorService`) intentionally skips the config update on that PUT and re-provisions the previously saved config. Any config modifications in that apply were silently discarded: the PUT returned 200, the apply succeeded, and nothing told the user their change was ignored. The same modify via the Portal UI (no `provision` param) was rejected with a validation error.

## Why a warning, not an error

The ticket asks for the UI's validation error to be surfaced on the Terraform side. We deliberately emit a **warning** instead: the PUT with `provision=true` on a FAILED resource is the supported retry-by-reapply mechanism, and failing the apply would break provisioning retries from Terraform. The request still goes through (re-provisioning the saved config); the user is now told their config changes were skipped and to re-apply once `provision_state` is `SUCCESS`.

The warning is suppressed when the update itself errors, and does not fire on retry-only re-applies (no config changes) or when provision mode is off (that path already surfaces the backend validation error).

## Test Plan

- [x] `TestWarnOnFailedStateUpdate` — warning fires on FAILED + config changes and the wrapped update is still invoked; no warning on retry-only re-apply, healthy resource, or provision off; suppressed when the update errors; appended alongside the update's own warnings
- [x] `go test ./alkira/` — full package suite passes
- [x] `go build ./...` and `go vet ./alkira/` clean

[AK-69357]: https://alkiranet.atlassian.net/browse/AK-69357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


snehachandrak:
One potential miss: resource_alkira_policy_nat.go

This resource has a provision_state schema attribute, sets it in CRUD operations, and checks client.Provision with provState == "FAILED" handling — the same pattern as the 55 wrapped resources. But it's not wrapped with warnOnFailedStateUpdate in this PR.

It's also missing the CustomizeDiff that forces provision_state from FAILED to SUCCESS (which triggers the update path on re-apply). This may be a pre-existing gap or intentionally different, but it's worth verifying whether policy_nat can end up in FAILED state and exhibit the same silent-skip behavior.

------
confirm if this change is applicable to the following Peering gateway / scale options resources

- peering_gateway_aws_tgw
- peering_gateway_aws_tgw_attachment
- peering_gateway_azure_vnet_third_party_connector_attachment
- peering_gateway_cxp
- network_entity_scale_options

use a "state" field with their own FAILED handling but no CustomizeDiff. The warnOnFailedStateUpdate wrapper reads from "provision_state" so it wouldn't work for them. These are correctly excluded from this PR, but they may have the same silent-skip problem — worth a follow-up ticket if so.

camjrichards:
Thanks @snehachandrak — verified against `tenant-provisioning-service`. The silent-skip-on-FAILED logic isn't unique to `AbstractConnectorService`; the same check lives in `AbstractNetworkCRUDService.update()` (the `isProvisioningRequested(filters) && previousEntity instanceof DiffableNetworkEntity && ProvisionedState.FAILED` branch that re-provisions the saved config and discards the request body). So the test is: does the resource's TPS service extend one of those two base classes and carry a `ProvisionedState`?

Findings:

- **policy_nat** — affected. `NATPolicyServiceImpl extends AbstractNetworkCRUDService` and the policy can enter `FAILED`. **Fixed in this PR** — added the `warnOnFailedStateUpdate` wrapper + the `provision_state` FAILED→SUCCESS `CustomizeDiff`, same as the other resources.

- **network_entity_scale_options** — also genuinely affected (`NetworkEntityScaleOptionsServiceImpl extends AbstractNetworkCRUDService`, model is a `DiffableNetworkEntity` with a `ProvisionedState`). It's *not* in the same bucket as the peering gateways. It stores its provision state in a field named `state` rather than `provision_state`, so the wrapper needs a small generalization to cover it. **Handling this in a follow-up ticket.**

- **peering_gateway_aws_tgw / aws_tgw_attachment / azure_vnet_third_party_connector_attachment / cxp** — correctly excluded, no change needed (here or follow-up). These use standalone services (`TransitGatewayServiceImpl`, etc.) that don't extend either base class, and their `state` is a lifecycle enum (`ACTIVE`/`REQUESTED`/`PENDING_ACCEPTANCE`/...), not a provisioning `ProvisionedState`. There's no `provision=true` skip path, so no silent-skip problem.

snehachandrak (approved):

PR #706, opened 2026-06-17
AK-69497: expose customer_end_overlay_ip on alkira_connector_ipsec_adv

rohanpillai-ak:
## Summary

Fixes [AK-69497](https://alkiranet.atlassian.net/browse/AK-69497). The `customer_end_overlay_ip` field on `alkira_connector_ipsec_adv` (tunnel block) was declared `Computed: true` only in the provider schema and was never read in `expandConnectorAdvIPSecTunnel`, so users could not set it from Terraform configuration. The underlying API, SDK, and backend validation already support the field as a writable override (`tenant-provisioning-service` ships `CustomerEndOverlayIPValidator` enforcing format, range, and /32-reservation rules).

## Changes

- `alkira/resource_alkira_connector_ipsec_adv.go` — `customer_end_overlay_ip` is now `Optional+Computed` (mirrors `prefix` on `alkira_ip_reservation`). Description spells out the override constraints.
- `alkira/resource_alkira_connector_ipsec_adv_helper.go` — `expandConnectorAdvIPSecTunnel` reads `customer_end_overlay_ip` from the tunnel config and populates `ConnectorAdvIPSecTunnelCustomerEnd.OverlayIp`.
- `examples/resources/alkira_connector_ipsec_adv/resource.tf` — added a second example block demonstrating the override.
- `docs/resources/connector_ipsec_adv.md` — moved the field from `Read-Only` to `Optional` with the override constraints written out.

## Test plan

- [ ] `terraform apply` on `alkira_connector_ipsec_adv` with `customer_end_overlay_ip = "<bare IPv4>"` set in a tunnel block succeeds when the customer-end overlay reservation is /32 and the IP does not overlap reserved ranges
- [ ] Omitting `customer_end_overlay_ip` continues to work — backend computes the value and `Read` populates state without drift
- [ ] Existing configurations without `customer_end_overlay_ip` produce no plan diff after upgrading the provider

## Related cleanup

The original ticket investigation went down the wrong path on the `IPReservation.customerEndIp` field. Reverts and closures:
- api-docs alkiranet/api-docs#302 (reverts #301, restores `readOnly: true` on `IPReservation.customerEndIp`)
- alkira-client-go alkiranet/alkira-client-go#253 (reverts #252, removes `CustomerEndIp` from `IPReservation` struct)
- tenant-provisioning-service alkiranet/tenant-provisioning-service#5356 — closed
- resource-manager alkiranet/resource-manager#1633 — closed
- terraform-provider-alkira #702 — closed; superseded by this PR

[AK-69497]: https://alkiranet.atlassian.net/browse/AK-69497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alkira/resource_alkira_connector_ipsec_adv.go
```diff
@@ -203,15 +203,38 @@ func resourceAlkiraConnectorIPSecAdv() *schema.Resource {
 									},
 									"customer_end_overlay_ip": {
 										Description: "The overlay IP address of " +
-											"the customer end of the tunnel.",
+											"the customer end of the tunnel. " +
```
snehachandrak, line 254: Consider enforcing the mutual exclusivity with` customer_end_overlay_ip_reservation_id ` in `CustomizeDiff` so that `plan` fails as well

rohanpillai-ak: Done — commit `2637fc35` wires it into `CustomizeDiff` using `d.GetRawConfig()` (rather than `d.Get()`) to avoid a false positive: `customer_end_overlay_ip` is Optional+Computed, so on subsequent plans of a reservation-only resource `d.Get` returns the backend-populated overlay IP alongside the user-set reservation id, which would trip the check. The raw-config walk only inspects fields the user literally wrote in HCL — mirrors the pattern in `resource_alkira_internet_applications.go`.

Verified on dev: `terraform plan` with both fields set now fails immediately with `tunnel customer_end_overlay_ip and customer_end_overlay_ip_reservation_id are mutually exclusive — supply one or the other. When customer_end_overlay_ip is set the cxp-end overlay reservation must be /32; otherwise use customer_end_overlay_ip_reservation_id.`

Existing override-only / reservation-only plans show no false-positive drift.

docs/resources/connector_ipsec_adv.md
snehachandrak, line 152: this can move to optional too, generate docs with `make doc`

rohanpillai-ak: Done — `customer_end_overlay_ip_reservation_id` is now under `Optional` and the description matches the strict either/or contract (commit `c36f7aef`).

Installed `tfplugindocs` (`go install github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs@latest`) and ran `make doc`. While I was there I also fixed the override example in `examples/resources/alkira_connector_ipsec_adv/resource.tf` — it had both `customer_end_overlay_ip` and `customer_end_overlay_ip_reservation_id` set, which is now rejected by both CustomizeDiff and the TPS validator. Updated to use only the override.

(I limited the commit to the AK-69497-related file; `make doc` also touched 83 unrelated doc files with whitespace-only changes from pre-existing drift, which I reverted to keep the PR scoped.)

snehachandrak (approved):

PR #635, opened 2026-04-24
Revert AWS VPC connector routing Read changes (#539, #588, #610)

asim-ak:
## Summary

Restores the AWS VPC connector resource to its pre-January-2026 behavior by reverting:
- **#539** (AK-65432): Read to populate routing and TGW fields
- **#588** (AK-66346): `Computed: true` + unconditional-set drift fix
- **#610** (AK-66870): `vpc_cidr` to `vpc_subnet` mode switch fix

## What changes

- Removes `Computed: true` from `vpc_cidr`, `vpc_subnet`, `vpc_route_table`, `overlay_subnets`
- Removes `setAwsVpcRoutingOptions` / `setAwsVpcExportPrefixes` / `setAwsVpcImportRouteTables` helper functions
- Read function no longer sets routing fields from the API response
- Restores original `expandUserInputPrefixes` logic

PR #712, opened 2026-06-29
AK-65432: revert AWS VPC connector routing Read changes (rel65)

snehachandrak:
Backport of dev revert #635 onto `release/2026-05-01-65`.

Reverts AK-65432/#539, AK-66346/#588, AK-66870/#610 — these introduced AWS VPC connector routing-Read drift regressions. Restores pre-Jan-2026 behavior (Read does not set routing fields; no `Computed` on those schema fields).

`go build ./...` and `go test ./alkira/...` pass.

### DIFF
--- a/alkira/resource_alkira_connector_aws_vpc.go
+++ b/alkira/resource_alkira_connector_aws_vpc.go
@@ -219,7 +219,6 @@ func resourceAlkiraConnectorAwsVpc() *schema.Resource {
 					},
 				},
 				Optional: true,
-				Computed: true,
 			},
 			"scale_group_id": {
 				Description: "The ID of the scale group associated with the connector.",
@@ -230,7 +229,6 @@ func resourceAlkiraConnectorAwsVpc() *schema.Resource {
 				Description: "Overlay subnet.",
 				Type:        schema.TypeList,
 				Optional:    true,
-				Computed:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
 			"description": {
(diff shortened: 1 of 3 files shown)


sankeerthmogili (commented):
Backport matches dev revert #635 byte-for-byte; the routing-Read drift fix and `Computed` removal look clean and self-consistent (removed helpers have no remaining references, unused `encoding/json` import dropped).

One item to confirm before merge: the revert re-introduces a `break` in the second loop of `setTgwAttachment` (the loop appending API-side TGW attachments not present in config). With that `break`, at most one out-of-band attachment is appended per Read, which can leave drift undetected when a user has >1.

Neither #588 nor #610 touched `setTgwAttachment`, and the current `release/2026-05-01-65` form of that loop has no `break` — so this may be over-reverting an unrelated, correct fix that already lives on the release branch. Could you confirm dropping the `break` is intended (faithful to #635) vs. preserving the no-`break` release-branch form?

snehachandrak:
Thanks for the detailed trace. I checked this across the release lineage, and the `break` is the long-standing behavior, not something this PR introduces:

- **v1.5.0 (released tag): has the `break`** — and `setTgwAttachment` there is byte-identical to this branch.
- **main: has the `break`. dev: has the `break`.**
- rel65 base is the *only* place without it — it came from #539 (commit `2989331f`), which removed the `break` but never reached the released line.

This backport restores the `break`, bringing rel65 in line with v1.5.0, main, and dev. So we're not adding the multi-attachment under-reporting — it's the existing shipped behavior, and keeping it is intentional (the goal here is rel65↔dev parity; dropping it would fork rel65 from all three).

The >1-untracked-attachment under-reporting is a real but pre-existing issue across v1.5.0/main/dev. Better handled as a dedicated fix on dev (then backported) than by diverging in this revert.

No change to this PR.

sankeerthmogili (approved):

sankeerthmogili (approved):

PR #726, opened 2026-07-01
AK-70237: Infoblox NIOS-X terraform support

ramakanthg:
Terraform provider support for NIOS-X on the Infoblox service (story AK-69565).

## Changes
- Instance `platform` (`NIOS`|`NIOS-X`, empty ⇒ NIOS); `model`/`type`/`password` relaxed **Required → Optional** (N/A for NIOS-X; back-compat — NIOS still provides them).
- NIOS-X auto-creates a real **`INFOBLOX_JOIN_TOKEN`** credential (was an `INFOBLOX_INSTANCE` password overload).
- Service `size` now accepts **`2LARGE`** (the `ValidateFunc` previously rejected it, though the API supports it).
- Example: a mixed **NIOS + NIOS-X** service; regenerated `docs/resources/service_infoblox.md`.
- Re-vendored local `alkira-client-go` (Platform field + join-token credential type).

## Testing
`go build ./...` + `go test ./alkira/ -run Infoblox` — green.

## Dependency note
Uses `alkira-client-go` (AK-70234) via the `replace ../alkira-client-go` + vendored copy (co-dev). Bump to the released client-go version + drop the replace once AK-70234 merges.


ramakanthg:
GitGuardian finding (id 34449477) was the example `join_token` — a fake placeholder (`eyJ0b2tlbiI6ImV4YW1wbGUifQ` = base64 of `{"token":"example"}`), not a real secret, so no rotation needed. Replaced it with `REPLACE_WITH_JOIN_TOKEN` in the example + generated doc so the scan is clean and the example doesn't look like a working token. Safe to dismiss the incident.

alkira/resource_alkira_service_infoblox.go
```diff
@@ -171,22 +176,45 @@ func resourceAlkiraInfoblox() *schema.Resource {
...
+						},
+						"platform": {
+							Description: "The platform type of the Infoblox " +
+								"instance. The value could be `NIOS` or " +
+								"`NIOS-X`. When not specified, it defaults to " +
+								"`NIOS`.",
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"NIOS", "NIOS-X"}, false),
```
sankeerthmogili: Decision: the API enum value is standardizing on `NIOS_X` (underscore) — please update to `StringInSlice([]string{"NIOS", "NIOS_X"}, false)` and change the other `"NIOS-X"` value literals to match (the `p != "NIOS-X"` check in this file and `r.Platform == "NIOS-X"` in the helper), plus docs/examples where the attribute value appears. Prose references to the product name can keep the hyphen. Mirrors client-side of TPS #5469 / orchestrator #2205 / api specs #495/#306.

go.mod
```diff
@@ -62,3 +62,5 @@ require (
 	k8s.io/klog/v2 v2.60.1 // indirect
 	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
 )
+
+replace github.com/alkiranet/alkira-client-go => ../alkira-client-go
```
sankeerthmogili: Merge-gate: the provider should consume alkira-client-go from the repo — not a local `replace` with hand-updated vendor files. alkiranet/alkira-client-go#254 is approved; once it merges, please run `go get github.com/alkiranet/alkira-client-go@main` here (picks up the merge as a pseudo-version — no release tag needed), drop this `replace`, and regenerate the vendor tree with `go mod vendor` (never edit vendor/ manually).

alkira/resource_alkira_service_infoblox.go
```diff
@@ -171,22 +176,45 @@ func resourceAlkiraInfoblox() *schema.Resource {
...
 						"password": {
 							Description: "The password associated with the " +
-								"infoblox instance.",
+								"infoblox instance. Not used for `NIOS-X` " +
+								"platform instances.",
 							Type:     schema.TypeString,
-							Required: true,
+							Optional: true,
+						},
+						"platform": {
```
sankeerthmogili, line 236: `platform` looks editable here, but it isn't — TPS rejects changing it after provisioning (both directions). As written, a customer who edits it gets a clean plan and a confusing apply-time failure.

Please surface it at plan time: a CustomizeDiff reject is the best fit (precedent: `policy_inter_cxp_routing`, which documents "immutability enforced in CustomizeDiff, not ForceNew") — `ForceNew` on an instance attribute would replace the whole service. At minimum, add "Immutable once the service is provisioned." to the description (pan precedent: `resource_alkira_service_pan.go`).

alkira/resource_alkira_service_infoblox_helper.go
```diff
@@ -47,6 +48,12 @@ func expandInfobloxInstances(in []interface{}, m interface{}) ([]alkira.Infoblox
 		if v, ok := instanceCfg["password"].(string); ok {
 			password = v
 		}
+		if v, ok := instanceCfg["platform"].(string); ok {
```
sankeerthmogili, line 78: (Anchoring here — the affected lines just above aren't in the diff.) **`expandInfobloxInstances` can delete the wrong instance.** The `instanceCfg["id"]` mapping at the top of this loop is *positional*: remove instance #2 of 3 and every id shifts down a slot, so the request tells TPS instance #3 is the one to remove. Not theoretical — dev testing watched the tail instance get destroyed twice instead of the removed one (finding F6). The `credential_id` mapping below has the same hazard.

Bluecat hit this exact bug and fixed it in this repo: `expandBluecatInstances(in, oldInstances, m)` re-maps `id`/credential ids **by hostname from old state** (`bluecat_helper.go:33-77`), guarded by a plan-time duplicate-hostname check (`validateBluecatInstanceHostnames`). Please copy that pattern — the doc-string advice to "append at the end of the list" papers over it.

ramakanthg:
Review items addressed in `c0a9c377`:

1. **NIOS_X rename**: `StringInSlice({"NIOS","NIOS_X"})`, the `p != "NIOS_X"` check, the helper comparison, error texts, example, and doc value references. Prose keeps the product name.
2. **Plan-time platform immutability**: CustomizeDiff rejects a platform change on an existing service, matched by hostname (not ForceNew — that would replace the whole service). Description updated.
3. **Positional id shift (F6)**: `expandInfobloxInstances` now takes old state and re-maps `id` **and** `credential_id` by hostname (Bluecat pattern); a new instance whose list slot inherited another instance's credential_id gets a fresh credential instead. `validateInfobloxInstanceHostnames` rejects duplicate hostnames at plan time. Unit tests added for the remove-middle-instance scenario.
4. **client-go merge-gate**: deferred until alkiranet/alkira-client-go#254 merges and a release is tagged — then bump `require`, drop the `replace`, `go mod vendor`. Tracked as a checklist item on this task.

ramakanthg:
Merge-gate resolved in `a6bea268`: alkira-client-go #254 is merged (0c295f61), so dropped the local `replace github.com/alkiranet/alkira-client-go => ../alkira-client-go`, pinned `require` to the released main pseudo-version (`v1.59.1-0.20260707001840-0c295f61322e`), and regenerated `vendor/` via `go mod vendor` (no hand edits — only go.mod/go.sum/modules.txt changed; the vendored source was already identical to the merged commit). `go build ./...` + infoblox unit tests green. No more `replace` in go.mod.

docs/resources/service_infoblox.md
```diff
@@ -1,5 +1,5 @@
 ---
-page_title: "alkira_service_infoblox Resource - terraform-provider-alkira"
+page_title: "alkira_service_infoblox Resource - alkira"
```
sankeerthmogili: The regenerated doc's `page_title` says `- alkira`, but all 72 other docs in the repo use `- terraform-provider-alkira` (e.g. `service_bluecat.md`, `service_pan.md`) — looks like `tfplugindocs` was run with the provider name set to `alkira`. Please re-run the generator with the correct provider name (so it doesn't regress on the next regeneration) or fix this line.

While you're touching the branch: `gofmt -l` still flags `resource_alkira_service_infoblox_helper.go` and `resource_alkira_service_infoblox_helper_test.go` (the misaligned `"platform":` map keys) — `gofmt -w` on both would clear it.

ramakanthg:
Addressed in 8b2b3835:
- `page_title` → `terraform-provider-alkira` (only this one doc was off; root cause was a manual `tfplugindocs --provider-name alkira` run — the `GNUmakefile` target infers the name correctly, so `make docs` won't regress).
- `gofmt -w` on `resource_alkira_service_infoblox_helper.go` + `_test.go` (the misaligned `"platform":` map key). `gofmt -l` now clean for both.

(The other four comments — NIOS_X enum, client-go `@main` drop-replace, plan-time platform immutability via CustomizeDiff, and the hostname-remap for the positional instance-id bug — are already in `bcca39fc`/`f0ffd6a7`.)

sankeerthmogili (approved):

PR #733, opened 2026-07-13
AK-70045: reject additional_tunnels_per_node that conflicts with tunnel options

sankeerthmogili:
## Problem

`alkira_network_entity_scale_options` exposes `additional_tunnels_per_node` as an optional user value, but the API **derives** it from the number of `additional_tunnel_options_per_node` (labels) and always stores that count. When the configured value differs from the label count — including the zero value produced when the field is omitted — config and the API-stored value diverge and the resource drifts on **every** plan (AK-70045). Reproduced on a live Prisma SD-WAN connector.

## Why this approach

`additional_tunnels_per_node` must stay **customer input** (it is set directly when no labels are used), so `Computed` is out. Making the omit case transparently work would require `alkira-client-go` to serialize the field as `*int` + `omitempty` (it is currently a plain `int`, always sent) **plus** `Computed` on the schema — cross-repo surgery for marginal gain. A `CustomizeDiff` guard is self-contained, keeps the field as customer input, and mirrors the server contract.

## Fix

`CustomizeDiff`: when `additional_tunnel_options_per_node` is set, require `additional_tunnels_per_node` to equal the label count; otherwise fail the plan with:

```
segment_scale_options[0]: additional_tunnels_per_node (2) must equal the number of additional_tunnel_options_per_node (1)
```

Logic extracted into `validateAdditionalTunnelsPerNodeMatchesTunnelOptions` for direct unit testing.

## Behavior

| Config | Result |
|--------|--------|
| labels + `additional_tunnels_per_node` == count | valid |
| labels + `additional_tunnels_per_node` != count | plan error |
| labels + `additional_tunnels_per_node` omitted (0) | plan error |
| no labels + `additional_tunnels_per_node` = N | unchanged (honored) |

## Testing

- Unit test `TestValidateAdditionalTunnelsPerNodeMatchesTunnelOptions` (matching / mismatched / omitted / no-labels / empty / multi-segment). Verified it fails without the guard.
- Live verification against a Prisma SD-WAN connector: mismatch → clear plan error; matched value → `No changes`.

## Related / rollout

- Server-side counterpart: tenant-provisioning-service PR #5499 (stops silently overwriting `additionalTunnelsPerNode`; validates instead).
- **Rollout ordering:** the released provider sends `additionalTunnelsPerNode: 0` when the field is omitted; once TPS #5499 rejects mismatches, this provider fix should ship with/ahead of it so label users are guided at plan time rather than hitting a 400.
- No `alkira-client-go` change required.

Note on acceptance-test-prod (expected transient failure)
Prod has not yet deployed AK-69179, so it still enforces the OLD policy and rejects the special character:

constraint.invalidpassword: Password should be minimum of 8 characters long ...
No special characters allowed.
The old-prod and new-preprod policies are mutually exclusive, so no single fixture value can pass both while prod lags. Since AK-69179 is not feature-flagged, once prod picks up that build it will enforce the same new policy and Ak123456789! will pass prod too.

Action for reviewer/merger: the acceptance-test-prod red is a known transient failure caused by the prod deployment lag, not by this change. It will clear on its own once prod deploys AK-69179; the prod check can be overridden to merge.

Jira: AK-70555

### DIFF
--- a/alkira/resource_alkira_network_entity_scale_options.go
+++ b/alkira/resource_alkira_network_entity_scale_options.go
@@ -29,6 +29,10 @@ func resourceAlkiraNetworkEntityScaleOptions() *schema.Resource {
 				d.SetNew("state", "SUCCESS")
 			}
 
+			if err := validateAdditionalTunnelsPerNodeMatchesTunnelOptions(d.Get("segment_scale_options").([]interface{})); err != nil {
+				return err
+			}
+
 			return nil
 		},
 		Importer: &schema.ResourceImporter{
@@ -138,6 +142,36 @@ func resourceAlkiraNetworkEntityScaleOptions() *schema.Resource {
 			},
 		},
 	}
+}
+
+// validateAdditionalTunnelsPerNodeMatchesTunnelOptions enforces that, when
+// additional_tunnel_options_per_node (labels) are set, additional_tunnels_per_node
+// equals the label count. The API derives additional_tunnels_per_node from the
+// number of labels, so a differing value - including the zero value produced when
+// the field is omitted - would diverge from the API-stored value and drift on
+// every plan (AK-70045). Failing here surfaces a clear plan-time error instead.
+func validateAdditionalTunnelsPerNodeMatchesTunnelOptions(segmentScaleOptions []interface{}) error {
+	for i, raw := range segmentScaleOptions {
+		sso, ok := raw.(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		labels, _ := sso["additional_tunnel_options_per_node"].([]interface{})
+		if len(labels) == 0 {
+			continue
+		}
+
+		tunnelsPerNode, _ := sso["additional_tunnels_per_node"].(int)
+		if tunnelsPerNode != len(labels) {
+			return fmt.Errorf(
+				"segment_scale_options[%d]: additional_tunnels_per_node (%d) must equal "+
+					"the number of additional_tunnel_options_per_node (%d)",
+				i, tunnelsPerNode, len(labels))
+		}
+	}
+
+	return nil
 }
 
 // convertEntityIdToInt converts entity_id string to int
--- a/alkira/resource_alkira_network_entity_scale_options.go
+++ b/alkira/resource_alkira_network_entity_scale_options.go
@@ -29,7 +29,7 @@ func resourceAlkiraNetworkEntityScaleOptions() *schema.Resource {
 				d.SetNew("state", "SUCCESS")
 			}
 
-			if err := validateAdditionalTunnelsPerNodeMatchesTunnelOptions(d.Get("segment_scale_options").([]interface{})); err != nil {
+			if err := validateAdditionalTunnelsPerNodeMatchesTunnelOptions(d.Get("segment_scale_options").([]any)); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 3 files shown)


alkira/resource_alkira_network_entity_scale_options.go
```diff
@@ -140,6 +144,36 @@ func resourceAlkiraNetworkEntityScaleOptions() *schema.Resource {
 	}
 }
 
+// validateAdditionalTunnelsPerNodeMatchesTunnelOptions enforces that, when
+// additional_tunnel_options_per_node (labels) are set, additional_tunnels_per_node
+// equals the label count. The API derives additional_tunnels_per_node from the
+// number of labels, so a differing value - including the zero value produced when
+// the field is omitted - would diverge from the API-stored value and drift on
+// every plan (AK-70045). Failing here surfaces a clear plan-time error instead.
+func validateAdditionalTunnelsPerNodeMatchesTunnelOptions(segmentScaleOptions []interface{}) error {
```
amitmandke: **Low — docs:** The constraint enforced here isn't surfaced anywhere a user reads before hitting the error. The schema `Description` for `additional_tunnels_per_node` (line ~74) still describes it as a free "Optional" value, and `docs/resources/network_entity_scale_options.md` is generated from that description, so neither mentions that it must equal the label count when `additional_tunnel_options_per_node` is set.

Consider appending to the schema `Description`, e.g.: `"When additional_tunnel_options_per_node (labels) are configured, this must equal the number of labels — the API derives it from the label count (AK-70045)."` then regenerate `docs/`. Guides label users at write time rather than at plan-failure time.

alkira/resource_alkira_network_entity_scale_options.go
```diff
@@ -140,6 +144,36 @@ func resourceAlkiraNetworkEntityScaleOptions() *schema.Resource {
...
+	for i, raw := range segmentScaleOptions {
+		sso, ok := raw.(map[string]interface{})
```
amitmandke: **Nit — style consistency:** This new code uses `[]interface{}` / `map[string]interface{}`, while the rest of this file uses the `any` alias (`[]any` / `map[string]any`, e.g. `generateNetworkEntityScaleOptionsRequest` and `resourceNetworkEntityScaleOptionsRead`). Functionally identical — align on `any` here (the call site on line 32, the func signature, and these assertions) for local consistency.

amitmandke (approved):
Approving. Correct, well-scoped fix for the AK-70045 drift bug: the CustomizeDiff guard forces config == label count for label users (config and API-stored value stay in sync, no drift) while leaving the field as free customer input when no labels are used. Rationale for choosing the guard over `Computed` (which would break the no-label input case) and over the `*int`+`omitempty` client change is sound — confirmed `AdditionalTunnelsPerNode` is a plain always-serialized `int` in alkira-client-go.

Verified:
- Unit test passes locally (`go test -run TestValidateAdditionalTunnelsPerNodeMatchesTunnelOptions` → ok); table covers matching / mismatched / omitted / no-labels / empty / multi-segment.
- Prior review comments both addressed: schema `Description` + regenerated docs now surface the constraint; production code switched to the `any` alias.
- Label counting uses `len(labels)` (all labels regardless of `enabled`) — confirmed with author this is intended; all labels are always counted.

Ship with or ahead of TPS #5499 per the rollout note.

snehachandrak (approved):

PR #737, opened 2026-07-16
AK-70005: add allow_list to alkira_connector_cisco_sdwan (cat8k-gated)

AK-Benjamin:
Add an optional allow_list attribute (Set of IPv4 CIDRs/IPs) to the Cisco SD-WAN connector resource for connector-level management-access control, in parity with Fortinet. Elements are validated by a new validateIPv4CidrOrIP helper: IPv4 address or IPv4 CIDR only (IPv6 is rejected, and host-bits-set CIDRs are rejected rather than silently masked), matching the backend, which is IPv4-only and requires the network address. Via CustomizeDiff, allow_list is only permitted on Cisco Catalyst 8000V (CAT8000V) connectors. Wire the value through read and request expansion (using the Set helper) and document the attribute via tfplugindocs.

Re-vendor alkira-client-go to pick up the ConnectorCiscoSdwan.AllowList field (AK-70004).

### DIFF
--- a/alkira/resource_alkira_connector_cisco_sdwan.go
+++ b/alkira/resource_alkira_connector_cisco_sdwan.go
@@ -28,6 +28,13 @@ func resourceAlkiraConnectorCiscoSdwan() *schema.Resource {
 				d.SetNew("provision_state", "SUCCESS")
 			}
 
+			// allow_list is only supported on Cisco Catalyst 8000V (cat8k)
+			// connectors. Fail fast at plan time for other types.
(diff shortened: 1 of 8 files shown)


sankeerthmogili (approved):

sankeerthmogili:
Nit (non-blocking, can be a follow-up): consider moving `validateIPv4CidrOrIP` from `resource_alkira_connector_cisco_sdwan_helper.go` into the shared `alkira/helper.go`. It's the provider's first IP/CIDR validator, and the other SD-WAN connectors (and the existing unvalidated CIDR fields like `policy_prefix_list`, `aws_vpc`, `oci_vcn`) will want to reuse it rather than each hand-rolling their own.

PR #745, opened 2026-08-04
AK-71184:make aws_account_id optional for AWS VPC connector

yashgupta912121:
## Summary

  - Make `aws_account_id` optional for `alkira_connector_aws_vpc` resource. TPS now auto-detects
    the VPC owner from AWS when not provided.
  - Add `omitempty` to `VpcOwnerId` JSON tag so an empty string is not sent in the API payload
    when the field is omitted.

  ## Context

  A customer (cyberdefense) provided the wrong `aws_account_id` in their Terraform config, which
  caused provisioning to fail with a confusing "Transit Gateway not found" error. TPS now validates
  `vpcOwnerId` against the actual VPC owner from AWS and rejects mismatches at create time
  (see TPS PR for AK-71184).

  Since TPS can auto-detect the VPC owner from the AWS `DescribeVpcs` API, there's no reason to
  force users to provide it. Making it optional reduces misconfiguration risk.

  ## Changes

  - **`resource_alkira_connector_aws_vpc.go`**: Changed `aws_account_id` from `Required: true` to
    `Optional: true, Computed: true`. `Computed` ensures Terraform picks up the auto-populated value
    from TPS on read.
  - **`vendor/.../connector_aws_vpc.go`**: Added `omitempty` to `VpcOwnerId` JSON tag so the field
    is omitted from the request body when empty.

  ## Backward Compatibility

  - Existing configs that provide `aws_account_id` continue to work as before.
  - New configs can omit `aws_account_id` and TPS will auto-populate it.

  ## Test Plan

  - [x] `go build ./...` passes
  - [ ] Apply existing Terraform config with `aws_account_id` set — should work as before
  - [ ] Apply new config without `aws_account_id` — TPS should auto-detect and populate it
  - [ ] Apply config with wrong `aws_account_id` — TPS should reject with validation error

### DIFF
--- a/alkira/resource_alkira_connector_aws_vpc.go
+++ b/alkira/resource_alkira_connector_aws_vpc.go
@@ -35,9 +35,10 @@ func resourceAlkiraConnectorAwsVpc() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"aws_account_id": {
-				Description: "AWS Account ID.",
+				Description: "AWS Account ID. If not provided, it will be automatically detected from the VPC.",
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
+				Computed:    true,
 			},
 			"aws_region": {
 				Description: "AWS Region where VPC resides.",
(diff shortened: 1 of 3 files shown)


asim-ak:
@yashgupta912121 please regenerate the docs.

asim-ak (approved):

PR #748, opened 2026-08-10
AK-71042: add allow_list to alkira_connector_juniper_sdwan

yashgupta912121:
## Summary
- Add optional `allow_list` attribute (`TypeSet` of `String`) to `alkira_connector_juniper_sdwan` resource for management-access IP/CIDR whitelisting
- Add plan-time validation via `validateIPv4CidrOrIP` (rejects IPv6, garbage input, host-bits-set CIDRs) — matches the Cisco cat8k `allow_list` pattern
- Vendor updated `alkira-client-go` with `AllowList` field on `ConnectorJuniperSdwan` struct
- Regenerate docs and add unit tests

## Changes
- `alkira/resource_alkira_connector_juniper_sdwan.go`: Schema (`allow_list`), read path (`d.Set`), expand path (`convertTypeSetToStringList`)
- `alkira/resource_alkira_connector_juniper_sdwan_test.go`: New — schema, validator, and expand tests
- `docs/resources/connector_juniper_sdwan.md`: Regenerated via `make doc`
- `go.mod`, `go.sum`, `vendor/`: Updated `alkira-client-go` dependency

## Backward Compatibility
- Existing configs without `allow_list` are unaffected (`omitempty` omits the field from the request body)
- No `CustomizeDiff` type-gate needed — Juniper SSR is single-mode

## Test plan
- [x] `go build ./...` compiles cleanly
- [x] `go test ./alkira/... -run TestJuniperSdwan` — all 3 test cases pass
- [x] `make doc` regenerates docs with new attribute

### DIFF
--- a/alkira/resource_alkira_connector_juniper_sdwan.go
+++ b/alkira/resource_alkira_connector_juniper_sdwan.go
@@ -33,6 +33,17 @@ func resourceAlkiraConnectorJuniperSdwan() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"allow_list": {
+				Description: "Management-access allow-list of IPv4 CIDRs or " +
+					"IP addresses. When set, only these sources can reach " +
+					"the management interface of the connector instances.",
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateIPv4CidrOrIP,
+				},
+			},
 			"name": {
 				Description: "The name of the connector.",
 				Type:        schema.TypeString,
@@ -273,6 +284,7 @@ func resourceConnectorJuniperSdwanRead(ctx context.Context, d *schema.ResourceDa
 		}}
 	}
 
+	d.Set("allow_list", connector.AllowList)
 	d.Set("billing_tag_ids", connector.BillingTags)
 	d.Set("cxp", connector.Cxp)
 	d.Set("group", connector.Group)
@@ -415,6 +427,7 @@ func generateConnectorJuniperSdwanRequest(d *schema.ResourceData, m any) (*alkir
 
 	// Construct the request payload
 	connector := &alkira.ConnectorJuniperSdwan{
+		AllowList:             convertTypeSetToStringList(d.Get("allow_list").(*schema.Set)),
 		BillingTags:           convertTypeSetToIntList(d.Get("billing_tag_ids").(*schema.Set)),
 		Instances:             instances,
 		JuniperSsrVrfMappings: expandJuniperSdwanVrfMappings(d.Get("juniper_ssr_vrf_mapping").(*schema.Set)),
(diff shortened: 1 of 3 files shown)


asim-ak (approved):
Approved.

Verified locally on the PR head (worktree off `pull/748/head`):

- `go build ./...`, `go vet ./alkira/...`, `gofmt -l alkira/` — all clean
- full `go test ./alkira/...` suite green (`ok ... 106.061s`), including the new `TestJuniperSdwanAllowList*` tests
- `tfplugindocs generate --provider-name terraform-provider-alkira` produces no PR-relevant diff for `docs/resources/connector_juniper_sdwan.md`; the 67-file doc churn it reports also reproduces on a clean `dev` worktree, so it's pre-existing tfplugindocs-version drift, not this PR
- merges cleanly onto `dev`, and cleanly alongside #749/#750/#751 (all four touch `go.mod`, `go.sum`, `vendor/modules.txt`); the combined 4-way merge also builds
- read/expand wiring matches the merged cisco reference (#737) exactly

Non-blocking notes (shared across #748–#751):

1. **Changelog.** `CLAUDE.md` lists schema additions as mandatory for `docs/changes/unreleased.md`. None of the four PRs add an entry (#737 missed it too). One combined entry covering all five SD-WAN connectors would settle it.

2. **Clearing the list.** `convertTypeSetToStringList` returns `nil` for an empty set (`alkira/helper.go:160`), and `AllowList` is tagged `json:"allowList,omitempty"`. So removing every `allow_list` entry from HCL sends a payload with no `allowList` key at all. If the backend treats an absent key as "leave unchanged", a user can never clear an allow-list once set — they'd get a perpetual diff. Same behaviour as the already-shipped cisco and fortinet implementations, so not a regression from this PR, but worth confirming the PUT semantics; if absent means unchanged, this needs an explicit empty array rather than `omitempty`.

3. **No type gate.** Cisco gates `allow_list` on `type == "CAT8000V"` in `CustomizeDiff`. These connectors have no equivalent type discriminator, so the absence of a gate looks right — just confirming the backend accepts `allowList` for every instance flavour of this connector, not a subset.

4. **Test duplication.** `TestXxxAllowListValidator` and `TestXxxAllowListExpand` are byte-identical across all four PRs plus `resource_alkira_connector_cisco_sdwan_test.go`. They exercise shared helpers (`validateIPv4CidrOrIP`, `convertTypeSetToStringList`), not resource-specific behaviour, so one test next to the helpers would replace five copies. The schema test is the part that's genuinely per-resource and worth keeping duplicated.

5. **Vendor bump riders.** The client-go bump carries two changes unrelated to `allow_list`: `ConnectorAwsVpc.VpcOwnerId` loses `omitempty` (harmless — the provider always populates it from the required `aws_account_id` at `resource_alkira_connector_aws_vpc_helper.go:215`), and `AllowList` is added to `ServicePan`/`ServiceFortinet`/`ServiceCheckpoint` (not yet consumed by the provider). Worth a line in the PR body so it isn't mistaken for scope creep later.

6. **Consistency.** `alkira_connector_fortinet_sdwan.allow_list` is still `TypeList` with no element validation (`resource_alkira_connector_fortinet_sdwan.go:70`) while these five are `TypeSet` + `validateIPv4CidrOrIP`. Separate cleanup, but it's now the odd one out.

PR #758, opened 2026-08-12
AK-71054: add allow_list to alkira_service_pan

yashgupta912121:
## Summary

Adds an optional `allow_list` attribute to `alkira_service_pan`, whitelisting source IPs/CIDRs that may reach the firewall's management surface. Part of AK-70969 (self-service mgmt-access allow-list across all third-party SD-WANs and firewalls), in parity with the five connector resources already merged to `dev`.

`ServicePan.AllowList` (`json:"allowList,omitempty"`) from AK-71036 is already vendored, so no re-vendor is needed here.

## Changes

- **`alkira/resource_alkira_service_pan.go`** — `allow_list` schema attribute (`TypeSet`, `Optional`, String elem with per-element validation); `d.Set("allow_list", pan.AllowList)` in `resourceServicePanRead`.
- **`alkira/resource_alkira_service_pan_helper.go`** — `AllowList: convertTypeSetToStringList(...)` in `generateServicePanRequest` (Set path, not `convertTypeListToStringList`, which would panic on a Set).
- **`alkira/resource_alkira_service_pan_test.go`** — table-driven unit tests (below).
- **`docs/resources/service_pan.md`** — new attribute line.

**No type-gate.** PAN is single-mode and not sub-scoped, so unlike Cisco cat8k there is no `CAT8000V`-style gate to add. The resource's existing `CustomizeDiff` closure is unchanged, and a non-empty `allow_list` is accepted regardless of `type`/`version`.

The ticket says to use `validation.IsCIDR` "matching the Cisco cat8k allow_list already on dev" — but cat8k on `dev` actually uses `validateIPv4CidrOrIP`, and that helper was hoisted from `resource_alkira_connector_cisco_sdwan_helper.go` into `alkira/helper.go` in d59a526d precisely so the sibling tickets could share it. All five merged siblings (aruba, versa, vmware, juniper, prisma) use it.

|---|---|---|

Using `IsCIDR` would make PAN the odd one out in the parent story. Happy to switch if the ticket's wording was deliberate.

`make doc` regenerates the entire docs tree: it rewrites ~80 unrelated files, deletes `docs/guides/release-notes-v1.4.3.md` and `v1.4.4.md`, and in this file *removes* the hand-written "Strata Cloud Manager (SCM) Example Usage" prose, replacing it with comments pulled from the example `.tf`. I reverted all of that and kept only the `allow_list` line, matching the 1-line doc diff in each sibling PR.

I did verify the shortcut: I re-ran `make doc`, extracted the generated `allow_list` line, and diffed it against the committed one — byte-identical, differing only in line number (151 vs 144) due to the pre-existing drift.

## Pre-existing issue found (out of scope, needs its own ticket)

`docs/resources/service_pan.md` on `dev` is missing `routing_type`, `scm_enabled`, and `scm_folder` — they exist in the schema but were never regenerated into the docs. Related: the SCM prose was hand-edited into the `.md` rather than into `templates/`, so **`make doc` is currently destructive** for whoever runs it next. Worth a follow-up to move that prose into the template and document the three attributes.

## Testing

Table-driven unit tests using testify + `schema.TestResourceDataRaw` / `schema.NewSet`, no live backend (the repo has no enabled acceptance suite — `testAccPreCheck` is commented out in `provider_test.go`):

- `TestAlkiraServicePanAllowListSchema` — asserts `TypeSet`, `Optional`, String `Elem` carrying a `ValidateFunc`.
- `TestAlkiraServicePanAllowListValidator` — accepts IPv4 CIDRs and bare IPv4; rejects malformed, out-of-range, IPv6, IPv4-mapped-IPv6, and host-bits-set CIDRs.
- `TestAlkiraServicePanAllowListExpand` — populated set, duplicate collapsing, and empty set (no panic, no entries).
- `TestAlkiraServicePanAllowListGenerateRequest` — drives the real `generateServicePanRequest` path via `schema.TestResourceDataRaw` and asserts `AllowList` on the resulting `ServicePan`, across populated / absent / explicitly-empty (absent and empty both yield no entries, so `omitempty` drops the field).
- `TestAlkiraServicePanAllowListRead` — `resourceServicePanRead` against a mock client round-trips `allow_list` from the API response.

Verified locally:

- `go build ./...` — clean
- `go vet ./alkira/` — clean

`golangci-lint` is not installed on this machine, so the repo lint target was not run — CI will cover it.

### DIFF
--- a/alkira/resource_alkira_service_pan.go
+++ b/alkira/resource_alkira_service_pan.go
@@ -71,6 +71,17 @@ func resourceAlkiraServicePan() *schema.Resource {
 			StateContext: importWithReadValidation(resourceServicePanRead),
 		},
 		Schema: map[string]*schema.Schema{
+			"allow_list": {
+				Description: "Management-access allow-list of IPv4 CIDRs or " +
+					"IP addresses. When set, only these sources can reach " +
+					"the management interface of the service instances.",
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateIPv4CidrOrIP,
+				},
+			},
 			"billing_tag_ids": {
 				Description: "Billing tags to be associated with " +
 					"the resource. (see resource `alkira_billing_tag`).",
@@ -557,6 +568,7 @@ func resourceServicePanRead(ctx context.Context, d *schema.ResourceData, m inter
 		}}
 	}
 
+	d.Set("allow_list", pan.AllowList)
 	d.Set("billing_tag_ids", pan.BillingTagIds)
 	d.Set("bundle", pan.Bundle)
 	d.Set("cxp", pan.CXP)
--- a/alkira/resource_alkira_service_pan_helper.go
+++ b/alkira/resource_alkira_service_pan_helper.go
@@ -459,6 +459,7 @@ func generateServicePanRequest(d *schema.ResourceData, m interface{}) (*alkira.S
 	}
 
 	service := &alkira.ServicePan{
+		AllowList:                   convertTypeSetToStringList(d.Get("allow_list").(*schema.Set)),
 		BillingTagIds:               convertTypeSetToIntList(d.Get("billing_tag_ids").(*schema.Set)),
 		Bundle:                      d.Get("bundle").(string),
 		CXP:                         d.Get("cxp").(string),
--- a/alkira/resource_alkira_service_pan_test.go
+++ b/alkira/resource_alkira_service_pan_test.go
@@ -1,6 +1,7 @@
 package alkira
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"testing"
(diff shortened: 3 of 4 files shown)


AK-Benjamin (approved):

amitmandke (approved):
**Approved.** Reviewed together with #759 and #760 as one story (AK-70969), and against `features/AK-70969/design/design-doc.md` in `eng-design-specs`.

Nothing needs to change in this diff before merge. It matches the five merged siblings precisely: `TypeSet` + Optional + per-element validation, Set expand path, set-on-read, no type-gate — which is exactly what the design's § *Terraform changes* and § *Why no type-gate* call for.

### On your question 1 — `validateIPv4CidrOrIP` vs `validation.IsCIDR`: keep what you have

Your reasoning holds, and on stronger grounds than the PR body claims. I checked the TPS side rather than inferring it: `@CIDR` → `CIDRValidator` → `IPAndMask.isValid(String)` (`IPAndMask.java:102`) rebuilds the input as `networkAddress/mask` and requires string equality, so `10.0.0.5/24` is **rejected**; `SubnetUtils` is IPv4-only, so IPv6 is **rejected**. That is precisely `validateIPv4CidrOrIP`'s behaviour.

So the ticket's `validation.IsCIDR` and the design doc's `validation.Any(IsCIDR, IsIPAddress)` would both have accepted `10.0.0.5/24` at plan time and then taken a 400 from TPS at apply time. Your choice is the only one of the three that mirrors the backend contract, and it keeps PAN consistent with cat8k (`resource_alkira_connector_cisco_sdwan.go:219`) and all five merged siblings.

The residue is a doc bug, not a code bug: the design doc still names a validator nobody implemented. Worth a one-line correction in `eng-design-specs` so QA and the remaining TPS subtasks don't build against it.

### On your question 2 — the single generated docs line: fine

Matches the 1-line doc diff in every sibling PR, and you verified it byte-identical against the generator. The real problem is that `make doc` is currently destructive (the SCM prose was hand-edited into `docs/` instead of `templates/`), which is what forces the shortcut. Agreed that deserves its own ticket — and #759 and #760 each found another instance (`service_fortinet.md` missing `alkira_admin_password`; `service_checkpoint.md` still calling `management_server` a `Block Set`). One docs-sync ticket covering all three plus the template move would unblock `make doc` for everyone.

On TPS `dev` (`00e7170bf`) `PANFWService.java` has no `allowList` field, and it carries `@JsonIgnoreProperties(ignoreUnknown = true)` (`:53`) with `FAIL_ON_UNKNOWN_PROPERTIES` globally disabled (`WebMvcConfig.java:196`). AK-71052 (TPS) and AK-71053 (Orch) are both still **Open**.

If a provider release ships before those land, a user who sets `allow_list` gets no error: TPS silently drops the field, Read returns nothing, and **every subsequent `plan` shows the same non-empty diff that never converges** — which also breaks story AC8. Merging to `dev` is not releasing, and VMware (#749) / Versa (#751) already merged under the same condition, so this is the story's accepted ordering rather than anything wrong here. Worth writing down on AK-70969 so the gate is explicit across nine subtasks: no provider release ahead of the per-target TPS/Orch tasks and their `RELEASE-SERVICE-*-ALLOW-LIST` flags.

### Optional nit

`TestAlkiraServicePanAllowListRead` uses `require.False(t, diags.HasError())`. #760 makes the sharper choice — `require.Empty(t, diags)` — because a failed GET here surfaces as a `diag.Warning` (`resource_alkira_service_pan.go:553`), which `HasError()` would not catch. Not a hole in practice, since the following `ElementsMatch` on the round-tripped value would fail anyway. Purely a consistency point if you touch the file again.

### Verified

`go build ./...` clean, `go vet ./alkira/` clean, new tests pass on `65c1654`. Create and update both route through `generateServicePanRequest` (`:495`, `:635`), so the field flows on update too. The `instance` test key matches the schema (`:188`). `CustomizeDiff` untouched and unaffected. Vendored `ServicePan.AllowList` confirmed present.

PR #759, opened 2026-08-12
AK-71057: add allow_list to alkira_service_fortinet

yashgupta912121:
## Summary

Adds an optional `allow_list` attribute to `alkira_service_fortinet`, whitelisting source IPs/CIDRs that may reach the firewall's management surface. Part of AK-70969 (self-service mgmt-access allow-list across all third-party SD-WANs and firewalls), following the same shape as the five merged connector resources and #758 (AK-71054, `alkira_service_pan`).

`ServiceFortinet.AllowList` (`json:"allowList,omitempty"`) from AK-71036 is already vendored, so no re-vendor is needed here.

## Changes

- **`alkira/resource_alkira_service_fortinet.go`** — `allow_list` schema attribute (`TypeSet`, `Optional`, String elem with per-element validation); `d.Set("allow_list", f.AllowList)` in `resourceFortinetRead`.
- **`alkira/resource_alkira_service_fortinet_helper.go`** — `AllowList: convertTypeSetToStringList(...)` in `generateFortinetRequest` (Set path, not `convertTypeListToStringList`, which would panic on a Set).
- **`alkira/resource_alkira_service_fortinet_test.go`** — table-driven unit tests (below).
- **`docs/resources/service_fortinet.md`** — new attribute line.

**No type-gate.** Fortinet FW is single-mode and not sub-scoped, so unlike Cisco cat8k there is no `CAT8000V`-style gate to add. The resource's existing `CustomizeDiff` closure is unchanged.

Both are identical to #758, so a decision there applies here too.

The ticket says to use `validation.IsCIDR` "matching the Cisco cat8k allow_list already on dev" — but cat8k on `dev` actually uses `validateIPv4CidrOrIP`, and that helper was hoisted into `alkira/helper.go` in d59a526d precisely so the sibling tickets could share it. All five merged siblings (aruba, versa, vmware, juniper, prisma) use it.

|---|---|---|

Using `IsCIDR` would make the two firewall resources inconsistent with the six connectors. Happy to switch if the ticket's wording was deliberate.

`make doc` regenerates the entire docs tree: ~80 unrelated files rewritten plus deletion of `docs/guides/release-notes-v1.4.3.md` and `v1.4.4.md`. I kept only the `allow_list` line, matching the 1-line doc diff in each sibling PR, and verified it by re-running `make doc` and diffing — byte-identical to the generator's output for this attribute.

## Pre-existing issue found (out of scope)

Regenerating showed `docs/resources/service_fortinet.md` on `dev` is also missing **`alkira_admin_password`** — it exists in the schema (and is `Sensitive`) but was never regenerated into the docs. Same class of drift as the `scm_*` attributes I flagged on `service_pan.md` in #758; worth folding both into one docs-sync follow-up.

## Testing

Table-driven unit tests using testify + `schema.TestResourceDataRaw` / `schema.NewSet`, no live backend (the repo has no enabled acceptance suite — `testAccPreCheck` is commented out in `provider_test.go`):

- `TestAlkiraServiceFortinetAllowListSchema` — asserts `TypeSet`, `Optional`, String `Elem` carrying a `ValidateFunc`.
- `TestAlkiraServiceFortinetAllowListValidator` — accepts IPv4 CIDRs and bare IPv4; rejects malformed, out-of-range, IPv6, IPv4-mapped-IPv6, and host-bits-set CIDRs.
- `TestAlkiraServiceFortinetAllowListExpand` — populated set, duplicate collapsing, and empty set (no panic, no entries).
- `TestAlkiraServiceFortinetAllowListGenerateRequest` — drives the real `generateFortinetRequest` path via `schema.TestResourceDataRaw` and asserts `AllowList` on the resulting `ServiceFortinet`, across populated / absent / explicitly-empty (absent and empty both yield no entries, so `omitempty` drops the field). Uses a mock client for the management-server segment lookup and a pre-set instance `credential_id` to avoid credential creation.
- `TestAlkiraServiceFortinetAllowListRead` — `resourceFortinetRead` against a mock client round-trips `allow_list` from the API response.

Note: the pre-existing `TestFortinetRead` in this file is `t.Skip`-ped ("mock server response format doesn't match API client expectations"). My read test does not depend on that path — it exercises `resourceFortinetRead` with a `ManagementServer`-less payload, so it runs for real rather than being skipped.

Verified locally:

- `go build ./...` — clean
- `go vet ./alkira/` — clean

`golangci-lint` is not installed on this machine, so the repo lint target was not run — CI will cover it.

### DIFF
--- a/alkira/resource_alkira_service_fortinet.go
+++ b/alkira/resource_alkira_service_fortinet.go
@@ -34,6 +34,17 @@ func resourceAlkiraServiceFortinet() *schema.Resource {
 			StateContext: importWithReadValidation(resourceFortinetRead),
 		},
 		Schema: map[string]*schema.Schema{
+			"allow_list": {
+				Description: "Management-access allow-list of IPv4 CIDRs or " +
+					"IP addresses. When set, only these sources can reach " +
+					"the management interface of the service instances.",
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateIPv4CidrOrIP,
+				},
+			},
 			"auto_scale": {
 				Description: "Whether enable auto scale for Fortinet firewall. " +
 					"It could be either `ON` and `OFF`. Default value is `OFF`.",
@@ -345,6 +356,7 @@ func resourceFortinetRead(ctx context.Context, d *schema.ResourceData, m interfa
 		}}
 	}
 
+	d.Set("allow_list", f.AllowList)
 	d.Set("auto_scale", f.AutoScale)
 	d.Set("billing_tag_ids", f.BillingTags)
 	d.Set("credential_id", f.CredentialId)
--- a/alkira/resource_alkira_service_fortinet_helper.go
+++ b/alkira/resource_alkira_service_fortinet_helper.go
@@ -293,6 +293,7 @@ func generateFortinetRequest(d *schema.ResourceData, m interface{}) (*alkira.Ser
 	}
 
 	service := &alkira.ServiceFortinet{
+		AllowList:        convertTypeSetToStringList(d.Get("allow_list").(*schema.Set)),
 		AutoScale:        d.Get("auto_scale").(string),
 		BillingTags:      billingTagIds,
 		CredentialId:     d.Get("credential_id").(string),
--- a/alkira/resource_alkira_service_fortinet_test.go
+++ b/alkira/resource_alkira_service_fortinet_test.go
@@ -1,7 +1,15 @@
 package alkira
 
 import (
+	"context"
+	"encoding/json"
+	"net/http"
 	"testing"
+
+	"github.com/alkiranet/alkira-client-go/alkira"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestFortinetRead(t *testing.T) {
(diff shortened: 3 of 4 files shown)


AK-Benjamin (dismissed):

amitmandke (approved):
**Approved.** Reviewed together with #758 and #760 as one story (AK-70969), and against `features/AK-70969/design/design-doc.md` in `eng-design-specs`.

Nothing needs to change in this diff before merge. Conforms to the merged sibling pattern and to the design's § *Terraform changes* / § *Why no type-gate*.

### On your question 1 — `validateIPv4CidrOrIP` vs `validation.IsCIDR`: keep what you have

Confirmed against the TPS side rather than inferred: `@CIDR` → `CIDRValidator` → `IPAndMask.isValid(String)` (`IPAndMask.java:102`) rebuilds the input as `networkAddress/mask` and requires string equality, so `10.0.0.5/24` is **rejected**; `SubnetUtils` is IPv4-only, so IPv6 is **rejected**. That is exactly `validateIPv4CidrOrIP`.

Both the ticket's `validation.IsCIDR` and the design doc's `validation.Any(IsCIDR, IsIPAddress)` would have accepted `10.0.0.5/24` at plan time and then taken a 400 at apply time. Your choice is the only one that matches the backend contract, and it keeps the firewall resources consistent with cat8k and the five merged connectors. The design doc is the thing that needs correcting, not the code.

### On your question 2 — the single generated docs line: fine

Matches every sibling PR, and you verified it against the generator's output. You've now found the third instance of the same drift (`alkira_admin_password` here, `scm_*` in #758, `management_server` block type in #760) — all downstream of `make doc` being destructive because hand-written prose lives in `docs/` instead of `templates/`. One docs-sync ticket covering the template move plus all three files is the right fix, and it isn't this story's work.

On TPS `dev` (`00e7170bf`), `model/fortinetfwservice/FortinetFWService.java` has no `allowList` field and carries `@JsonIgnoreProperties(ignoreUnknown = true)` (`:51`), with `FAIL_ON_UNKNOWN_PROPERTIES` globally disabled (`WebMvcConfig.java:196`). AK-71055 (TPS) and AK-71056 (Orch) are both still **Open**.

If a provider release ships first, a user who sets `allow_list` gets no error — TPS drops the field silently, Read returns nothing, and **every subsequent `plan` shows the same non-empty diff that never converges**, breaking story AC8. Merging to `dev` isn't releasing, and VMware (#749) / Versa (#751) merged under the same condition, so this is the story's ordering rather than a problem with this PR. Worth making the gate explicit on AK-70969: no provider release ahead of the per-target TPS/Orch tasks and their `RELEASE-SERVICE-*-ALLOW-LIST` flags.

### Optional nit

`TestAlkiraServiceFortinetAllowListRead` uses `require.False(t, diags.HasError())`; #760 uses `require.Empty(t, diags)` and explains why — a failed GET here returns a `diag.Warning` (`resource_alkira_service_fortinet.go:341`) that `HasError()` won't catch. Not a real hole, since the `ElementsMatch` on the round-tripped value would fail regardless. Consistency only.

### Verified

`go build ./...` clean, `go vet ./alkira/` clean, new tests pass on `c182bcd`. Create and update both route through `generateFortinetRequest` (`:285`, `:410`). The `instances` test key is correct for this resource (`:97`) — worth calling out, since PAN and Check Point use the singular `instance` and getting it wrong would have silently emptied the test. `CustomizeDiff` untouched. Vendored `ServiceFortinet.AllowList` confirmed present. Also confirmed your new read test genuinely runs rather than inheriting the skip on `TestFortinetRead`.

amitmandke (dismissed):
**Approved** — re-review after the force-push. My earlier approval on `c182bcd` carries forward.

### The push was a pure rebase, not a content change

- `git diff c182bcd 6ab6f7ce -- 'alkira/*fortinet*' 'docs/resources/service_fortinet.md'` → **0 lines**. The Fortinet change is byte-identical to what I reviewed.
- The old head was based on `1ea4f953` (AK-71048, prisma); this head is based on `815f1883` — your merged #760 (AK-71060, checkpoint), which is the current `dev` tip. A naive old-head→new-head diff therefore *looks* like it adds the checkpoint code; that code is in the base now, not in this PR.
- Against the merge-base: 4 files, +175/−0 — exactly the diff I reviewed.

### Verified at `6ab6f7ce`

`go build ./...` clean, `go vet ./alkira/` clean, `gofmt -l alkira/` empty. All 10 `*AllowList*` tests pass — the Fortinet set and the newly-merged checkpoint set together in the same package, no symbol collisions from the rebase (checkpoint's `newCheckpointAllowListMockClient` vs Fortinet's inline `createMockAlkiraClient`, no duplicate test names). All six CI checks green at this SHA.

### Follow-up on my earlier comments

- **Optional nit, still open** (`resource_alkira_service_fortinet_test.go:225`): `require.False(t, diags.HasError())` vs `require.Empty(t, diags)`. Now that #760 has merged with `require.Empty` plus the explanatory comment, this is the last one of the set on the weaker assertion. Still cosmetic — the `ElementsMatch` on the round-tripped value fails either way — and still optional. Not blocking; fold it in only if you touch the file again.
- **Release ordering (AK-70969)**: re-checked Jira — AK-71055 (TPS) and AK-71056 (Orch) are both still **Open**, and TPS `dev` still has no `allowList` on `FortinetFWService`. The "provider releases ahead of the backend → `plan` diff that never converges" risk is unchanged. Story-level ordering, not a blocker here, but please keep the gate explicit on AK-70969.
- Validator choice, the single-line docs diff, and the docs-sync follow-up ticket: all as agreed in my previous review, nothing to redo.

### One FYI, not a finding

`mergeStateStatus` reads `BLOCKED` even though the PR is `MERGEABLE`, `APPROVED`, and green on all checks. I can't read `dev`'s branch protection without repo admin, so I can't say which rule is holding it — worth a look from someone who can before assuming it's ready to land. Nothing in the diff.

asim-ak (approved):

PR #760, opened 2026-08-12
AK-71060: add allow_list to alkira_service_checkpoint

yashgupta912121:
## Summary

Adds an optional `allow_list` attribute to `alkira_service_checkpoint`, whitelisting source IPs/CIDRs that may reach the firewall's management surface. Part of AK-70969, following the same shape as the five merged connector resources, #758 (`alkira_service_pan`) and #759 (`alkira_service_fortinet`).

`ServiceCheckpoint.AllowList` (`json:"allowList,omitempty"`) from AK-71036 is already vendored, so no re-vendor is needed here.

## Changes

- **`alkira/resource_alkira_service_checkpoint.go`** — `allow_list` schema attribute (`TypeSet`, `Optional`, String elem with per-element validation); `d.Set("allow_list", checkpoint.AllowList)` in `resourceCheckpointRead`.
- **`alkira/resource_alkira_service_checkpoint_helper.go`** — `AllowList: convertTypeSetToStringList(...)` in `generateCheckpointRequest` (Set path, not `convertTypeListToStringList`, which would panic on a Set).
- **`alkira/resource_alkira_service_checkpoint_test.go`** — table-driven unit tests (below).
- **`docs/resources/service_checkpoint.md`** — new attribute line.

**No type-gate.** Check Point FW is single-mode and not sub-scoped, so unlike Cisco cat8k there is no `CAT8000V`-style gate to add. The resource's existing `CustomizeDiff` closure is unchanged.

Both are identical to #758 and #759, so one decision covers all three.

The ticket says to use `validation.IsCIDR` "matching the Cisco cat8k allow_list already on dev" — but cat8k on `dev` actually uses `validateIPv4CidrOrIP`, and that helper was hoisted into `alkira/helper.go` in d59a526d precisely so the sibling tickets could share it. All five merged siblings use it.

|---|---|---|

`make doc` regenerates the entire docs tree: ~80 unrelated files rewritten plus deletion of two release-notes guides. I kept only the `allow_list` line, matching the sibling PRs, and verified it by re-running `make doc` and diffing — byte-identical to the generator's output for this attribute.

## Testing

Table-driven unit tests using testify + `schema.TestResourceDataRaw` / `schema.NewSet`, no live backend:

- `TestAlkiraServiceCheckpointAllowListSchema` — asserts `TypeSet`, `Optional`, String `Elem` carrying a `ValidateFunc`.
- `TestAlkiraServiceCheckpointAllowListValidator` — accepts IPv4 CIDRs and bare IPv4; rejects malformed, out-of-range, IPv6, IPv4-mapped-IPv6, and host-bits-set CIDRs.
- `TestAlkiraServiceCheckpointAllowListExpand` — populated set, duplicate collapsing, and empty set (no panic, no entries).
- `TestAlkiraServiceCheckpointAllowListGenerateRequest` — drives the real `generateCheckpointRequest` path and asserts `AllowList` on the resulting `ServiceCheckpoint`, across populated / absent / explicitly-empty.
- `TestAlkiraServiceCheckpointAllowListRead` — `resourceCheckpointRead` against a mock client round-trips `allow_list` from the API response.

### Note on the read test — worth a look

Two traps here that are easy to get wrong, and both are why the pre-existing `TestCheckpointRead` in this file is `t.Skip`-ped ("mock server response format doesn't match API client expectations"):

1. **`resourceCheckpointRead` reports a failed GET as a `diag.Warning`, not an error.** A test asserting `require.False(diags.HasError())` therefore passes even when nothing was fetched and every `d.Set` was skipped. This test asserts the diagnostics are *empty* instead, which is what caught the second problem.
2. **Segment lookup by name unmarshals into `[]alkira.Segment`, but lookup by ID unmarshals into a single object.** The mock routes on the `name=` query parameter and returns the right shape for each. It also returns exactly one segment, since `resourceCheckpointRead` hard-fails on any other count.

The mock additionally serves instance-credential creation, so `generateCheckpointRequest` runs without a live backend.

Verified locally:

- `go build ./...` — clean
- `go vet ./alkira/` — clean

`golangci-lint` is not installed on this machine, so the repo lint target was not run — CI will cover it (it passed on #758 and #759).

## Pre-existing docs drift (out of scope, now seen on all three)

Regenerating showed `docs/resources/service_checkpoint.md` on `dev` also documents `management_server` as `Block Set, Min: 1` when the schema now says `Block List, Min: 1, Max: 1`. That is the third instance of stale docs in this family — `service_pan.md` is missing `routing_type`/`scm_enabled`/`scm_folder`, `service_fortinet.md` is missing `alkira_admin_password`. Worth one docs-sync ticket covering all three.

### DIFF
--- a/alkira/resource_alkira_service_checkpoint.go
+++ b/alkira/resource_alkira_service_checkpoint.go
@@ -34,6 +34,17 @@ func resourceAlkiraCheckpoint() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"allow_list": {
+				Description: "Management-access allow-list of IPv4 CIDRs or " +
+					"IP addresses. When set, only these sources can reach " +
+					"the management interface of the service instances.",
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateIPv4CidrOrIP,
+				},
+			},
 			"auto_scale": {
 				Description: "Indicate if `auto_scale` should be enabled " +
 					"for your checkpoint firewall. `ON` and `OFF` are " +
@@ -374,6 +385,7 @@ func resourceCheckpointRead(ctx context.Context, d *schema.ResourceData, m inter
 		return diag.FromErr(fmt.Errorf("failed to find segment"))
 	}
 
+	d.Set("allow_list", checkpoint.AllowList)
 	d.Set("auto_scale", checkpoint.AutoScale)
 	d.Set("billing_tag_ids", checkpoint.BillingTags)
 	d.Set("credential_id", checkpoint.CredentialId)
--- a/alkira/resource_alkira_service_checkpoint_helper.go
+++ b/alkira/resource_alkira_service_checkpoint_helper.go
@@ -309,6 +309,7 @@ func generateCheckpointRequest(d *schema.ResourceData, m interface{}) (*alkira.S
 
 	// Assemble request
 	return &alkira.ServiceCheckpoint{
+		AllowList:        convertTypeSetToStringList(d.Get("allow_list").(*schema.Set)),
 		AutoScale:        d.Get("auto_scale").(string),
 		BillingTags:      convertTypeSetToIntList(d.Get("billing_tag_ids").(*schema.Set)),
 		CredentialId:     d.Get("credential_id").(string),
--- a/alkira/resource_alkira_service_checkpoint_test.go
+++ b/alkira/resource_alkira_service_checkpoint_test.go
@@ -1,7 +1,16 @@
 package alkira
 
 import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"strings"
 	"testing"
+
+	"github.com/alkiranet/alkira-client-go/alkira"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCheckpointRead(t *testing.T) {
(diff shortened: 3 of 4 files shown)


AK-Benjamin (approved):

amitmandke (approved):
**Approved.** Reviewed together with #758 and #759 as one story (AK-70969), and against `features/AK-70969/design/design-doc.md` in `eng-design-specs`.

Nothing needs to change before merge. Conforms to the merged sibling pattern and to the design's § *Terraform changes* / § *Why no type-gate*.

### On the validator question (raised across all three PRs): keep `validateIPv4CidrOrIP`

Confirmed against TPS rather than inferred: `@CIDR` → `CIDRValidator` → `IPAndMask.isValid(String)` (`IPAndMask.java:102`) rebuilds the input as `networkAddress/mask` and requires string equality, so `10.0.0.5/24` is **rejected**; `SubnetUtils` is IPv4-only, so IPv6 is **rejected**. That is precisely `validateIPv4CidrOrIP`'s behaviour.

The ticket's `validation.IsCIDR` and the design doc's `validation.Any(IsCIDR, IsIPAddress)` would both have accepted `10.0.0.5/24` at plan time and then taken a 400 from TPS at apply time. Your choice is the only one of the three that matches the backend contract. The design doc needs a one-line correction; the code is right.

### On the read test — the note in your description is the best thing in this PR

Both traps you identified are real and I verified them:

1. `resourceCheckpointRead` returns `diag.Warning` on a failed GET, so `require.False(diags.HasError())` would pass with nothing fetched. `require.Empty(t, diags)` is the correct assertion, and it is the one #758 and #759 don't use — the same Warning shape exists in both (`resource_alkira_service_pan.go:553`, `resource_alkira_service_fortinet.go:341`). Neither is actually holed, because the following `ElementsMatch` on the round-tripped value would fail anyway, but yours is the assertion the other two should have used.
2. Segment lookup really does unmarshal into `[]alkira.Segment` by name and a single object by ID, and `resourceCheckpointRead` hard-fails on any count other than one. Routing the mock on the `name=` query parameter is the right way to handle it.

Writing both of those down rather than leaving them for the next person to rediscover is worth more than the diff itself.

On TPS `dev` (`00e7170bf`), `CheckPointFWService.java` has no `allowList` field and carries `@JsonIgnoreProperties(ignoreUnknown = true)` (`:46`), with `FAIL_ON_UNKNOWN_PROPERTIES` globally disabled (`WebMvcConfig.java:196`). AK-71058 (TPS) and its Orch pair are both still **Open**.

If a provider release ships before those land, a user who sets `allow_list` gets no error — TPS drops the field silently, Read returns nothing, and **every subsequent `plan` shows the same non-empty diff that never converges**, which also breaks story AC8. Merging to `dev` isn't releasing, and VMware (#749) / Versa (#751) merged under the same condition, so this is the story's accepted ordering. Worth making the gate explicit on AK-70969: no provider release ahead of the per-target TPS/Orch tasks and their `RELEASE-SERVICE-*-ALLOW-LIST` flags.

### Docs drift

Agreed on folding all three findings into one docs-sync ticket — `management_server` block type here, `alkira_admin_password` in #759, `scm_*` in #758. The root cause is shared: `make doc` is destructive because hand-written prose lives in `docs/` rather than `templates/`, which is what forces the 1-line shortcut in all six sibling PRs. The template move belongs in that ticket too.

### Verified

`go build ./...` clean, `go vet ./alkira/` clean, new tests pass on `9c4441b`. Create and update both route through `generateCheckpointRequest` (`:300`, `:416`). The `instance` test key matches the schema (`:78`). `CustomizeDiff` untouched. Vendored `ServiceCheckpoint.AllowList` confirmed present. One pre-existing quirk, unchanged by this PR and not worth acting on: `d.Set("allow_list", ...)` sits after the `numOfSegments != 1` early return, so a 0- or 2+-segment service never reaches it — but that is equally true of every other attribute in that function today.

PR #763, opened 2026-08-12
AK-71658: set inbound_connector_type in internet application Read

amitmandke:
## AK-71658 — IFA Terraform import leaves `inbound_connector_type` unset

Jira: https://alkiranet.atlassian.net/browse/AK-71658
Slack: https://alkiranet.slack.com/archives/C01J2UKNQBA/p1786559249746009

### Problem

`terraform import` of an `alkira_internet_application` produces a spurious in-place update on the very next `terraform plan`:

```
  # alkira_internet_application.ifa-app will be updated in-place
  ~ resource "alkira_internet_application" "ifa-app" {
        id                     = "306"
      + inbound_connector_type = "DEFAULT"
        name                   = "ifa-app"
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

This happens even though the API returns the field:

```json
{ "id": 306, "name": "ifa-app", "inboundConnectorType": "DEFAULT", ... }
```

### Root cause

`resourceInternetApplicationRead` never called `d.Set("inbound_connector_type", ...)`. The field was only ever read *out* of state to build create/update requests; the value coming back from the API was discarded.

`terraform import` populates state solely from Read, so the attribute landed in state as `null`. On the next plan the config value — supplied by the schema `Default: "DEFAULT"` even when written explicitly — compared as `null → "DEFAULT"` and rendered as an addition.

The schema `Default` does not help here: defaults apply to configuration, never to state written by Read. Ordinary create/apply flows don't expose it because SDKv2 carries the planned value into state after apply, which is also why a single `terraform apply` makes the diff go away.

### Change

- Add `d.Set("inbound_connector_type", app.InboundConnectorType)`.
- Move the direct API-response → state mappings out of `Read` into `setInternetApplicationFields`, so the mapping has a seam that unit tests can drive without an API client. Fields needing lookups or reshaping (segment, source NAT pool, targets, provision state) stay in `Read`.

### Tests

`TestInternetApplicationSetFields` — verified as a real regression test: deleting the new `d.Set` line makes both subtests fail (`expected "AKAMAI_PROLEXIC", actual ""`), restoring it makes them pass.

`gofmt`, `go vet`, and the full `go test ./alkira/` suite are clean.

### Out of scope (deliberately not fixed here)

Two related gaps in the same Read, left alone because neither is what AK-71658 reports and one cannot be fixed safely in this repo:

- **`inbound_connector_id`** is also never set. It is *not* a one-line fix: the API returns `inboundInternetConnectorId` (int) while `alkira-client-go` maps `InboundConnectorId string` to `inboundConnectorId`, a key absent from the response. Adding a `d.Set` today would write `""` on every refresh and *introduce* a perpetual diff for `AKAMAI_PROLEXIC` IFAs. Needs an `alkira-client-go` change first.
- **`group_id`** is `Computed` and never set. A null computed attribute produces no plan diff, so this is a data-completeness gap rather than a broken plan. The correct source field is ambiguous (`inboundInternetGroupId` vs `connectorGroupId`) and no other resource in the repo sets a `group_id`.

### Verification

Re-run the AK-71310 IFA import automation; `terraform plan` immediately after import should report no changes.


snehachandrak (dismissed):
Superseded — see inline comment.

alkira/resource_alkira_internet_applications.go
```diff
@@ -342,6 +342,30 @@ func resourceInternetApplicationCreate(ctx context.Context, d *schema.ResourceDa
...
+// lands in state as null and shows up as a spurious diff on the next plan.
+func setInternetApplicationFields(d *schema.ResourceData, app *alkira.InternetApplication) {
+	d.Set("billing_tag_ids", app.BillingTags)
+	d.Set("bi_directional_az", app.BiDirectionalAvailabilityZone)
+	d.Set("byoip_id", app.ByoipId)
+	d.Set("connector_id", app.ConnectorId)
+	d.Set("connector_type", app.ConnectorType)
+	d.Set("description", app.Description)
+	d.Set("fqdn_prefix", app.FqdnPrefix)
+	d.Set("inbound_connector_type", app.InboundConnectorType)
```
snehachandrak: Confirmed in TPS: `InternetApplication` is `@JsonInclude(NON_NULL)` and `InternetApplicationTranslator.toModel` passes the entity value straight through, so the field is **omitted** whenever `inbound_connector_type` is NULL (nullable column, no default). TPS treats NULL as `DEFAULT` (`InboundConnectorType.equals`), so mirror that rather than setting it blind:

```suggestion
	if app.InboundConnectorType != "" {
		d.Set("inbound_connector_type", app.InboundConnectorType)
	} else {
		d.Set("inbound_connector_type", "DEFAULT")
	}
```

As written, a NULL-column IFA lands `""` in state and diffs forever against the `DEFAULT` config — and import still wouldn't be fixed for those.

snehachandrak (changes requested):

alkira/resource_alkira_internet_applications.go
amitmandke: Good catch — confirmed independently and fixed in cb6850b0.

I walked the same path in TPS and it holds end to end: `InternetApplicationEntity` declares `@Column(name = "inbound_connector_type")` with no `nullable = false` and no default, and `InternetApplicationTranslator.toModel` calls `.inboundConnectorType(entity.getInboundConnectorType())` unconditionally — with Lombok `@Builder`, an explicit `.field(null)` overrides the model's `@Builder.Default` of `DEFAULT`. Combined with `@JsonInclude(NON_NULL)`, the key is omitted from the response, so the Go client sees `""`.

You're right that it would have been strictly worse than the status quo, not just an incomplete fix: today those IFAs land `null` and render as `+ DEFAULT`, which one apply clears. With the blind `d.Set` they'd land `""` and render as `~ "" -> "DEFAULT"` on every plan — and because Read rewrites it on each refresh, it would never converge.

I implemented it as a small named helper rather than the inline `if/else` so the reasoning has somewhere to live next to the behavior, but the semantics are exactly your suggestion:

```go
func inboundConnectorTypeOrDefault(t string) string {
	if t == "" {
		return "DEFAULT"
	}
	return t
}
```

Also added `omitted inbound_connector_type falls back to DEFAULT` as a regression test — worth noting the pre-existing subtests passed either way, so that new case is the one actually covering the NULL-column path. Verified by reverting the normalization: it fails with `expected DEFAULT, actual ""`.

One more thing I checked while in there: `internet_protocol` is the only other field in this resource with a schema `Default`, and it needs no equivalent handling — the same translator null-coalesces it to `IPV4` before serialization, so it is always present.

Branch is rebased on latest `dev` (it was 6 behind). PTAL.

snehachandrak (approved):

PR #765, opened 2026-08-13
AK-72272: stop exposing production portal credentials to pre-review PR builds

sankeerthmogili:
Story: AK-73369
Closes AK-72272

## What was wrong

`.github/workflows/acceptance-prod.yml` and `.github/workflows/acceptance-preprod.yml` both:

1. triggered on `pull_request` to `main`, `dev` and `release/*`,
2. exported `ALKIRA_PORTAL` / `ALKIRA_USERNAME` / `ALKIRA_PASSWORD` from the `PROD_TERRAFORM_*` and `PREPROD_TERRAFORM_*` secrets at **workflow** scope, with no `environment:` gate, then
3. ran `make` and `terraform apply --auto-approve` on the PR's **own checked-out code**.

So any pull request could read those live portal credentials — a one-line change to the Makefile is enough to print or exfiltrate them — and could mutate the production tenant, all before a reviewer saw the diff.

**Accurate scoping for review:** GitHub does not expose secrets to `pull_request` runs from forks, so this was never an anonymous-internet issue; the actor had to already have push access to open a branch PR. It is an insider / compromised-account exposure of live production credentials, plus an availability risk from unreviewed provider code applying against prod. That is why it is still worth closing.

## What changed

Removed the `pull_request:` trigger block from both workflows. `push:` to `main`/`dev`/`release/*` and `workflow_dispatch:` are unchanged, so:

- acceptance still runs on every merge to a protected branch (code that has been reviewed), and
- anyone who wants acceptance coverage on a branch can still get it via `workflow_dispatch`, as a deliberate act.

Nothing else in the workflows was touched.

## Backward compatibility

Backward-compatible. CI configuration only — no provider code, no schema, no state, no product behaviour.

Verified rather than assumed: neither `main` nor any repository ruleset requires an acceptance-test status check (`gh api repos/alkiranet/terraform-provider-alkira/rules/branches/main` returns `[]`, and there is no branch-protection object on `main`), so removing the trigger cannot leave a PR blocked waiting on a check that will never report.

## Deliberately not done here

- **No protected `environment:` added.** If the team does want acceptance on every PR, the idiomatic GitHub control is a protected `environment:` with required reviewers on the job, which holds the run until a human approves before secrets are injected. That is a repo-settings change as well as a workflow change, so it belongs in a follow-up rather than being half-implemented here. Either approach is fine; please do not add a bespoke gate.
- **Secrets left at workflow scope.** Scoping the `env:` block to the apply/destroy steps so the build step never sees them is worth doing, but each of these workflows has exactly one job, so it buys little today and would enlarge this diff.
- **Credentials not rotated.** These prod/preprod portal credentials have been reachable from every PR build for as long as the trigger has existed. Treat them as exposed and rotate them separately — this PR does not do that, and merging it does not make an already-leaked credential safe.
- **No move to a short-lived scoped API key.** The provider already prefers `api_key` and deprecates `password`; CI has not followed. Separate ticket.

## Release

CI-only, no user-visible provider change, but flagging it for the `review-terraform-release-notes` go/no-go anyway since it changes what runs before a release is cut.

### DIFF
--- a/.github/workflows/acceptance-prod.yml
+++ b/.github/workflows/acceptance-prod.yml
@@ -2,11 +2,10 @@ name: acceptance-test-prod
 
 on:
   workflow_dispatch:
-  pull_request:
-    branches:
-      - main
-      - dev
-      - 'release/*'
+  # NOTE: deliberately no `pull_request:` trigger. This workflow exports live
+  # production portal credentials at workflow scope and then builds and runs
+  # the checked-out code, so a `pull_request` run would hand those secrets to
+  # unreviewed code. Use `workflow_dispatch` to run acceptance on a branch.
   push:
     branches:
       - main
--- a/.github/workflows/acceptance-preprod.yml
+++ b/.github/workflows/acceptance-preprod.yml
@@ -2,11 +2,10 @@ name: acceptance-test-preprod
 
 on:
   workflow_dispatch:
-  pull_request:
-    branches:
-      - main
-      - dev
-      - 'release/*'
+  # NOTE: deliberately no `pull_request:` trigger. This workflow exports live
+  # preprod portal credentials at workflow scope and then builds and runs the
+  # checked-out code, so a `pull_request` run would hand those secrets to
+  # unreviewed code. Use `workflow_dispatch` to run acceptance on a branch.
   push:
     branches:
       - main


.github/workflows/acceptance-prod.yml
```diff
@@ -2,11 +2,10 @@ name: acceptance-test-prod
 
 on:
```
amitmandke, line 3: The residual the ticket's threat model actually describes isn't closed by removing the `pull_request` trigger.

`push:` to `dev` still exports prod portal creds and applies to the prod tenant, and there's **no branch protection or ruleset** on `main` or `dev` — I checked: `rules/branches/main` → `[]`, `branches/dev/protection` → 404, `environments` → `total_count: 0`. So the same "engineer with push access" who previously exfiltrated via a PR can push directly to `dev` and get the identical run.

The description cites the absent protection as evidence that no required check breaks. That's true, and it's equally evidence that "pre-review" isn't enforced after this fix.

Separately, `workflow_dispatch:` accepts an arbitrary `ref`, so any write-access actor can dispatch this workflow against their own unmerged branch and get the same secrets. The bar moves from automatic-and-silent-on-every-PR to deliberate-and-attributable, which is a real improvement — but the description frames dispatch as "a reviewer runs it once the code is trusted", and GitHub enforces no such gate. Only a protected `environment:` with required reviewers would.

amitmandke (approved):
**Approved.**

AK-72272 / VULN-006. Dropping the `pull_request` trigger is one of the two fix options the ticket itself proposes, and it's the simpler one.

**Verified empirically, not just by reading the YAML:** on this PR only `lint`, `unit-test` and `dependency-review` ran; on every other recent PR (`task/AK-71057`, `security/AK-72024-...`) both `acceptance-test-prod` and `acceptance-test-preprod` ran on `pull_request`. The removal takes effect from the merge commit.

**Swept the whole post-PR `.github/workflows/` tree**, not just the changed files: no `pull_request_target`, no `issue_comment`, no `workflow_run` anywhere, and no remaining `secrets.*` reference on any fork- or PR-reachable trigger. `lint.yaml`, `unit-test.yaml` and `dependency-review.yml` reference zero secrets; `build-branch`/`build-latest` are `push`-only; `build-release` is tag-only. The repo is public with forking enabled, so your scoping in the description is right — GitHub never handed secrets to fork PRs, this was always insider/compromised-account.

The CI signal you lose is real but loud, not silent: acceptance genuinely ran and genuinely caught breakage on PRs (`AK-71090-security-bumps` failed, `security/AK-72024-...` failed then passed). Post-change that signal moves to the `push: dev` run, so a regression surfaces after merge instead of before. Nothing silently green-checks and nothing blocks, since no required status check exists. Reasonable trade for closing the exposure.

Unrelated to this diff: `dependency-review` is red on pre-existing `go.mod` state (grpc 1.79.3 GHSA-hrxh-6v49-42gf, x/net 0.48.0).

Three notes below on what the change doesn't reach. None blocks the merge — the scanner finding is closed — but they matter for whether the ticket closes.

---

### Findings outside the changed lines
_These are in code this PR does not touch, so GitHub can't anchor them inline._

**`.github/workflows/acceptance-prod.yml:19`** — Minor, and you defer it explicitly in the description — agreed it's small given one job per workflow.

Secrets stay at workflow scope, so on `push`/`dispatch` runs the `make` build step still has `PROD_TERRAFORM_PASSWORD` in its environment.

.github/workflows/acceptance-prod.yml
sankeerthmogili: You're right, and I'll correct the framing rather than defend it. Removing the `pull_request` trigger narrows the surface from *automatic-and-silent on every PR* to *deliberate-and-attributable*, but it does **not** enforce pre-review: with no ruleset on `main`/`dev`, a push to `dev` still runs the job with prod creds, and `workflow_dispatch` accepts an arbitrary `ref`. The only thing that actually enforces "a reviewer runs it once the code is trusted" is a protected `environment:` with required reviewers gating the job. I'm treating that protected-environment gate as the real fix (follow-up, since it's a repo-admin change) and correcting the PR description to stop claiming dispatch is reviewer-gated. Thanks for checking the rulesets directly — the description shouldn't have leaned on their absence.

PR #766, opened 2026-08-13
AK-71090 Update google.golang.org/grpc and golang.org/x/net to address dependabot alerts (AK-70687)

asim-ak:
## AK-71090 / AK-70687 — grpc and x/net security bumps

Closes both AK-71090 and AK-70687, which cover the same two modules in this repo.

### Versions
Both are indirect dependencies, before and after:

| Module | Before | After |
|---|---|---|
| `google.golang.org/grpc` | v1.79.3 | v1.82.1 |
| `golang.org/x/net` | v0.48.0 | v0.55.0 |

`go mod tidy` pulled these along to satisfy the two bumps, all indirect: `x/mod` v0.30.0→v0.35.0, `x/sync` v0.19.0→v0.20.0, `x/sys` v0.39.0→v0.45.0, `x/text` v0.32.0→v0.37.0, `x/tools` v0.39.0→v0.44.0, `protobuf` v1.36.10→v1.36.11.

### Vendor diff
This repo vendors dependencies, so `go mod vendor` was required. The diff is large but mechanical — upstream restructured `x/net/idna` Unicode tables and reshuffled grpc internal packages.

One thing worth a reviewer's eye: vendoring also resynced `vendor/github.com/alkiranet/alkira-client-go/...` even though that module's pinned version in `go.mod` did not change. I checked this by vendoring fresh from the unmodified base commit in isolation, which produced identical content — so the committed `vendor/` tree on `main` was already out of sync with its own pinned version. That drift is pre-existing, not caused by this bump. Kept the resync in since it is a required side effect of vendoring correctly.

### Verification
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — `ok github.com/alkiranet/terraform-provider-alkira/alkira 108.8s`; root package has no test files

No tests skipped. The `acceptance/` directory holds standalone `.tf` configs for manual runs, not `go test` targets, and there are no `TF_ACC`-gated tests in the Go suite.


asim-ak:
Closing — this was raised against main. Reopening from dev instead, which is the correct base for this change. Superseding PR to follow.

PR #768, opened 2026-08-13
AK-71090 Update google.golang.org/grpc and golang.org/x/net to address dependabot alerts (AK-70687)

asim-ak:
## AK-71090 / AK-70687 — grpc and x/net security bumps

Closes both AK-71090 and AK-70687, which cover the same two modules in this repo.

Supersedes #766, which was raised against `main` by mistake. Rebuilt from `dev`.

### Versions
Both indirect dependencies, before and after:

| Module | Before | After |
|---|---|---|
| `google.golang.org/grpc` | v1.79.3 | v1.82.1 |
| `golang.org/x/net` | v0.48.0 | v0.55.0 |

`go mod tidy` pulled these along, all indirect: `x/mod` v0.30.0→v0.35.0, `x/sync` v0.19.0→v0.20.0, `x/sys` v0.39.0→v0.45.0, `x/text` v0.32.0→v0.37.0, `x/tools` v0.39.0→v0.44.0, `protobuf` v1.36.10→v1.36.11.

### Vendor diff
This repo vendors dependencies, so `go mod vendor` was required. The diff is large but mechanical — upstream restructured `x/net/idna` Unicode tables and reshuffled grpc internal packages.

The vendor tree is otherwise untouched: I verified by running a fresh `go mod vendor` from an unmodified `origin/dev` checkout in isolation and diffing. `vendor/github.com/alkiranet/alkira-client-go` came out byte-identical, so this diff is scoped purely to the two target modules and their transitives.

### Verification
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — `ok github.com/alkiranet/terraform-provider-alkira/alkira 105.9s`; root package has no test files

No tests skipped. The `acceptance/` directory holds standalone `.tf` configs for manual runs, not `go test` targets, and there are no `TF_ACC`-gated tests in the Go suite.


snehachandrak (changes requested):
## Review — Changes Requested (nothing verified this)

**Ticket:** AK-71090 — bump `google.golang.org/grpc` and `golang.org/x/net` to fix vulnerabilities.

The intent is right and the change is well scoped: a single commit, 10 commits behind `dev`, and the only non-vendor files are `go.mod` and `go.sum` with `vendor/` regenerated consistently. But **no CI ran on it**, and for a dependency bump that regenerates the vendor tree that is the whole problem.

### The blocker

The only check on this PR is **GitGuardian Security Checks — `skipping`**. There is no build, no `go vet`, no test run. So nothing has established that this tree even compiles, let alone behaves. That matters more here than for a typical bump:

- `golang.org/x/net v0.48.0 → v0.55.0` is **seven minor versions**. x/net moves `http2` and `proxy` internals between minors, and a Terraform provider talks to its API over HTTP/2 via gRPC — this is not a patch-level change being described as one.
- `google.golang.org/grpc v1.79.3 → v1.82.1` is three minors on the transport that the provider plugin protocol itself runs over.
- The vendor tree is regenerated (+20k/-72k), so a stale or partially-regenerated `vendor/` would fail at build time, which is exactly what CI would have caught.

**Please get a green build/vet/test run on this before merge.** If the workflow does not trigger for dependabot-style branches or for `vendor/`-only diffs, that is worth fixing separately — a bump nobody builds is not a bump anybody can trust.

### Two things to add to the description

- **Name the CVEs.** The title says "to fix vulnerabilities" without saying which. For an advisory-driven bump that is the one piece of information a reviewer and a release manager both need — it decides urgency, and it tells you whether `v0.55.0` is actually the minimum fixed version or just the latest.
- **These are all `// indirect`.** `x/net`, `grpc` and `protobuf` are every one of them indirect requirements. Bumping an indirect dependency by editing `go.mod` directly works, but it is not necessarily durable: the next `go mod tidy` can drop or re-resolve the requirement based on what the direct dependencies actually ask for. Worth confirming the bump survives a `go mod tidy` — otherwise the fix silently reverts on someone else's routine housekeeping commit, and the advisory reopens with the ticket closed. If a direct dependency is pinning the vulnerable version, the durable fix is bumping that one.

